PR #112, opened 2026-06-13
feat: new theme tests

eugenioenko:


eugenioenko:
These themes should go into the `/themes` distribution folder as part of #143 (install themes from GitHub repo) instead of being embedded in the binary.

PR #121, opened 2026-06-13
feat: add navigation history (back/forward)

### LINKED ISSUE #72: Navigation history — back/forward after go-to-definition
**Tier 1 — High Impact**

Alt+Left/Right to navigate back and forward after go-to-definition, find references, etc. Essential companion to existing LSP navigation.

Sources: VS Code, Zed, Sublime, HN, Reddit

eugenioenko:
## Summary
- Add `NavHistory` data structure in `internal/core/navhistory/` for browser-style back/forward navigation (push truncates forward history, max 100 entries with eviction)
- Register `navigate.back` (Alt+Left) and `navigate.forward` (Alt+Right) commands
- Push navigation history before: go-to-definition, go-to-implementation, go-to-type-definition, find references, go-to-line, search result navigation, and problems/references panel navigation
- Add Navigate Back/Forward to View menu and editor right-click context menu

## Test plan
- [x] Unit tests for `NavHistory` (push, back, forward, truncation, dedup, eviction, round-trip)
- [x] E2E tests: navigate back after go-to-line, forward after back, cross-file navigation, no-op on empty history, multiple go-to-line jumps
- [x] Functional test: go-to-line then Alt+Left back, Alt+Right forward, no-op with no history
- [x] `go build ./...` passes
- [x] `go test ./...` passes (all existing + new tests)

Closes #72


eugenioenko:
Navigation history is a must-have for an IDE, and the stack-based implementation is very clean. Some suggestions for broader integration:

1. Explorer Integration: Currently, navigation history seems to only be pushed during 'jumps' (LSP, Search). Consider also pushing to the history when opening a file from the Explorer, so users can easily jump back to their previous file.
2. Initial State: The current logic requires a 'priming' jump to start working. If a user starts at a file and wants to go back after their first jump, the first 'Back' action might not do anything because the initial position wasn't on the stack yet. Consider pushing the initial position on startup or first file open.

PR #127, opened 2026-06-13
feat: bracket pair colorization

### LINKED ISSUE #86: Bracket pair colorization
**Tier 3 — Differentiator**

Color-code nested bracket pairs for improved readability.

Sources: VS Code, Zed

eugenioenko:
## Summary
- Add bracket pair colorization that color-codes nested `()`, `[]`, `{}` pairs with depth-based cycling through 3 colors (gold, magenta, blue)
- Brackets inside strings and comments are excluded from colorization using syntax highlighting span data
- Add `editor.bracketPairColorization` setting (default `true`) and `view.toggleBracketColorization` command to toggle the feature

## Implementation
- New style constants `StyleBracket1`, `StyleBracket2`, `StyleBracket3` in `internal/term/screen.go`
- New `BracketPairColors` theme config struct in `internal/config/theme.go` with JSON-configurable colors
- Bracket color computation in `editor_widget.go` scans from buffer start to build correct nesting depth for visible range
- Colors are applied as foreground overrides during cell rendering, after syntax highlighting but before search highlights

## Test plan
- [x] E2E test: nested brackets `foo(bar(baz()))` get 3 different foreground colors
- [x] E2E test: closing brackets match their opening pair's color
- [x] E2E test: toggle command correctly enables/disables colorization
- [x] E2E test: brackets inside strings are NOT colorized
- [x] `go build ./...` passes
- [x] `go test ./...` passes (all existing + new tests)

Closes #86

### DIFF
--- a/docs-web/src/content/docs/guides/themes.md
+++ b/docs-web/src/content/docs/guides/themes.md
@@ -122,7 +122,7 @@ Below is a complete theme file showing every configurable section. All color val
       "bg": "#f92672",
       "fg": "#f8f8f2"
     },
-    "bracketColors": ["yellow", "magenta", "cyan"]
+    "bracketColors": ["yellow", "magenta", "blue"]
   },
   "menu": {
     "item": {},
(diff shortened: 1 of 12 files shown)


eugenioenko:
Bracket colorization is a great feature, but the current implementation has a significant performance bottleneck:

The computeBracketColors function scans and highlights the buffer from the very beginning up to the end of the visible range on every single render call. In large files (e.g., 10k+ lines), this will cause severe UI lag during scrolling and typing. Consider caching the nesting depth per line or only re-computing when the buffer actually changes, rather than on every frame.

PR #132, opened 2026-06-13
fix: dialog UX polish - auto-sizing, unicode hotkeys, hit regions

### LINKED ISSUE #122: UX polish: rendering and interaction improvements
## UX Polish Audit

A systematic review of the UI layer for rendering consistency, interaction rough edges, and small polish opportunities. These are not new features -- they are improvements to what already exists to make the editor feel more refined.

---

### Status Bar

- [ ] **Notification buttons lack visual differentiation from text** (`statusbar_widget.go:133-162`). The `[OK]`, `[Action]`, and `[Secondary]` buttons use the same style as the notification text. They should use a distinct style (e.g., bold or inverse) so users recognize them as clickable. Currently only brackets hint at interactivity.
- [ ] **Blame text can collide with right-aligned segments** (`statusbar_widget.go:55-56`). The blame text is rendered left-to-right without checking whether it overlaps the right-aligned position/encoding/language segments. A long blame string will visually clash with or overwrite the right side.
- [ ] **Status bar notifications have no auto-dismiss visual timer** (`statusbar.go:49-55`). Notifications with a `duration` silently expire. There is no visual countdown or fade to signal to the user how long the notification will persist, making timed notifications feel unpredictable.

### Tab Bar

- [ ] **Tab close button hit target is a single column** (`tabbar_widget.go:229`). The close button `x` on the active tab occupies exactly 1 column, making it very hard to click. Consider expanding the hit region to at least 3 columns (the `x` plus surrounding padding).
- [ ] **Inactive tabs show no close affordance** (`tabbar_widget.go:47-55`). Only the active tab renders the `x` close button. Users must first click a tab to select it, then click again to close -- two clicks instead of one. Consider showing `x` on hover or always.
- [ ] **No visual indication of dirty state on inactive tabs beyond the dot** (`tabbar_widget.go:134-151`). The `●` dirty marker on inactive tabs uses `StyleWarning` which may blend with the inactive tab style depending on the theme. Consider also bolding or underlining the tab name.
- [ ] **Right-click uses Button2 (middle click) instead of Button3** (`tabbar_widget.go:211`). `tcell.Button2` is the middle mouse button. Context menus should typically trigger on `tcell.Button3` (right click). However, this may be intentional for terminal compatibility -- verify.

### Editor

- [ ] **Magic number 3 for scroll wheel delta** (`editor_widget.go:566,578`). The scroll speed `3` is hardcoded in multiple places. This should be a named constant (e.g., `scrollWheelDelta`) or configurable.
- [ ] **Magic number 4 for horizontal scroll delta** (`editor_widget.go:565,576,587,590`). Similarly, horizontal scroll uses hardcoded `4`. Should be a named constant.
- [ ] **Magic number 400 for double-click timeout** (`editor_widget.go:650`). The double-click detection threshold is hardcoded at 400ms. Should be a named constant (e.g., `doubleClickTimeoutMs`).
- [ ] **No visual cursor blink** The editor cursor is a static block with no blink, which can make it hard to locate after scrolling. Consider supporting cursor blink as a setting.
- [ ] **Gutter fold chevron only appears on hover for expanded folds** (`editor_widget.go:251-254`). Collapsed folds show the `▶` always, but expanded folds only show `▼` when the gutter is being hovered. This asymmetry means users can't see fold indicators for expanded regions unless they hover.

### Find Bar / Replace Bar

- [ ] **Find bar hardcoded width of 40** (`findbar_widget.go:95`). The find bar width is fixed at 40 columns (or `sw-4` for small terminals). On wide terminals, 40 columns can feel cramped. Consider making it proportional to the editor width (e.g., 50% with a max).
- [ ] **Replace bar also hardcoded at 40 columns** (`replacebar_widget.go:112`). Same issue as find bar.
- [ ] **Find bar hardcoded Y position at 4** (`findbar_widget.go:100`). The find bar always appears at row 4 from the top of the editor, regardless of editor height. On very short terminals, this could overlap with other UI elements.
- [ ] **Navigation buttons (arrows) and close button are 1-cell hit targets** (`findbar_widget.go:147-157`). The `▲`, `▼`, and `✕` buttons each have a 1-cell click target, making them difficult to click precisely.
- [ ] **No visual feedback when no matches are found** (`findbar_widget.go:113`). When the search query has no matches, the counter shows `0/0` but there is no color change or "no matches" indicator to draw attention. Consider tinting the input border red or showing explicit "No results" text.

### Command Palette

- [ ] **Palette width hardcoded at 60** (`palette_widget.go:107`). The palette is always 60 columns wide. On wide terminals this could be wider, and on narrow terminals the `sw-4` fallback kicks in.
- [ ] **File list capped at 100 items** (`palette_widget.go:339,350`). The file browser caps at 100 results with no indication to the user that results are truncated. Consider showing a "100+ results, refine search" indicator.
- [ ] **File list capped at 10,000 entries** (`palette_widget.go:91`). For very large workspaces, the file walker stops at 10,000 files silently. Users searching for files beyond this limit will not find them.
- [ ] **No fuzzy matching** (`palette_widget.go:309,346`). Both command and file filtering use `strings.Contains` (substring matching). Fuzzy matching (matching by initials, e.g., "oc" matching "Open Changes") would significantly improve discoverability.

### Explorer (Sidebar)

- [ ] **No horizontal scroll for deeply nested or long file names** (`explorer_widget.go:229-235`). Long file names in deep directories are simply truncated with no way to see the full path. Consider showing a truncation indicator or horizontal scrolling.
- [ ] **No file type icons** (`explorer_widget.go:207-219`). Only folders get a chevron indicator. Files have no icon or indicator of their type, making it harder to scan the tree.
- [ ] **Active file highlight uses same style as selection** (`explorer_widget.go:193-194`). Both `Selected` and `ActiveFile` use `StyleSidebarSelected`, making it impossible to distinguish between "this is the file I'm editing" and "this is where my cursor is in the explorer."

### Dialogs

- [ ] **Confirm dialog minimum width of 30 is arbitrary** (`confirmdialog_widget.go:57`). The `boxW = 30` minimum can be too wide for short messages and too narrow for long ones. Consider auto-sizing based on content.
- [ ] **Input dialog button click regions are slightly off** (`inputdialog_widget.go:104-107`). The Save button boundary calculation uses `saveEnd` as `d.boxX + d.boxW - 2`, but the actual rendered text may differ by a column due to border. This could cause the rightmost pixel of the save button to be unclickable.
- [ ] **Confirm dialog hotkey matching is case-insensitive by adding 32** (`confirmdialog_widget.go:160`). The hotkey detection `rune(btn[0]+32)` only works for ASCII letters and would break for non-ASCII button labels. Should use `unicode.ToLower`.

### Context Menu

- [ ] **Context menu firstEvent guard can swallow legitimate clicks** (`contextmenu_widget.go:179-182`). The `firstEvent` flag discards the first mouse event to avoid the menu immediately responding to the click that opened it. However, this also means if a keyboard opens the menu and the user immediately clicks an item, the first click is lost.

### Search Widget

- [ ] **Search summary "X results in Y files" appears without padding** (`search_widget.go:577-580`). The summary line starts at column 0 with no left padding, while the input fields use the InputWidget prefix. This creates a visual misalignment.
- [ ] **"No results" and "Searching..." messages have no left padding** (`search_widget.go:549,559`). Same alignment issue as the summary -- these status messages start at column 0 while inputs are indented.
- [ ] **Replace/preview buttons (icons) on file rows are very small** (`search_widget.go:662-668`). The `⟳` and `⊙` replace/preview buttons are single Unicode characters at fixed positions, making them hard to discover and click.

### Scrollbar

- [ ] **Scrollbar uses same character for track and thumb** (`scrollbar.go:46-49`). Both use `'█'` but with different styles. In themes where these styles are similar, the scrollbar becomes invisible. Consider using different characters (e.g., `'░'` for track, `'█'` for thumb) or ensuring all themes have sufficient contrast.

### Autocomplete

- [ ] **All completion kinds use the same symbol** (`autocomplete_widget.go:26`). `CompletionKind.Symbol()` always returns `'■'` regardless of the kind. The kind differentiation relies solely on color. Using different symbols (e.g., `f` for function, `v` for variable, `T` for type) would improve accessibility and scanability.

### Hover Widget

- [ ] **Hover widget hardcoded max visible lines of 12** (`hover_widget.go:10`). The `hoverMaxVisibleLines = 12` is a hardcoded constant. On tall terminals, this could show more content, and on short terminals this might be too much.

### Terminal Panel

- [ ] **"No terminals" empty state message has no action hint for keyboard users** (`terminal_panel.go:117`). The message says "Press + to create one" which only applies to mouse users clicking the `+` button. Keyboard users need to know the keybinding (e.g., the shortcut for `terminal.new`).

### General / Cross-cutting

- [ ] **Inconsistent border set usage across widgets**. Some widgets default to `SingleBorderSet()` (context menu, find bar, hover), while the command palette defaults to `DoubleBorderSet()`. All should consistently use the configured border set from the app.
- [ ] **InputWidget prefix is inconsistent across contexts**. The default prefix is ` ❯ ` (with an emoji), but the command palette uses ` ` (space only), and the changes widget commit input has no prefix. This creates visual inconsistency.
- [ ] **No mouse hover feedback on clickable elements**. Clickable status bar segments (indent, EOL), tab close buttons, sidebar more button, find/replace navigation buttons -- none change appearance on hover to signal interactivity.
- [ ] **SelectableList PageUp/PageDown not supported** (`selectable_list.go`). The `SelectableList` only handles Up/Down/Enter keys. PageUp/PageDown would improve navigation in long lists (explorer, search results, problems).
- [ ] **Problems widget has no scrollbar** (`problems_widget.go`). Unlike the search widget which has a scrollbar, the problems panel has none, making it unclear how many items are off-screen.

eugenioenko:
## Summary
- **Confirm dialog auto-sizing**: Removed hardcoded `boxW = 30` minimum width. The dialog now auto-sizes based on `max(messageWidth, buttonsWidth) + padding`, clamped to a max of 60 columns (or screen width - 4 on narrow terminals)
- **Unicode-safe hotkey detection**: Replaced unsafe ASCII arithmetic `rune(btn[0]+32)` with `unicode.ToLower(rune(first[0]))` for proper Unicode support in button hotkeys
- **Input dialog hit regions**: Refactored click detection to store `HitRegion` structs during render and reuse them in the event handler, following the "single source of truth for layout" design principle and matching the pattern already used in `ConfirmDialogWidget`

Closes #122

## Test plan
- [x] Added `TestConfirmDialogUnicodeShortcut` - verifies Unicode button labels (e.g. "Offnen") match lowercase key presses
- [x] Added `TestConfirmDialogAutoSize` - verifies box width adapts to short and long messages
- [x] Added `TestConfirmDialogAutoSizeClamp` - verifies box clamps on narrow screens
- [x] Added `TestInputDialogHitRegionsStoredFromRender` - verifies save/cancel hit regions are correctly populated with valid positions
- [x] Added `TestInputDialogCustomConfirmLabel` - verifies custom confirm label affects hit region width
- [x] All existing tests pass (`go test ./...`)

### DIFF
--- a/internal/ui/confirmdialog_widget.go
+++ b/internal/ui/confirmdialog_widget.go
@@ -1,6 +1,8 @@
 package ui
 
 import (
+	"unicode"
+
 	"github.com/eugenioenko/ttt/internal/term"
 
 	"github.com/gdamore/tcell/v2"
@@ -8,13 +10,13 @@ import (
 
 type ConfirmDialogWidget struct {
 	BaseWidget
-	Message    string
-	Buttons    []string
-	Selected   int
-	Borders    *term.BorderSet
-	OnButton   []func()
-	OnDismiss  func()
-	btnHits []HitRegion
+	Message   string
+	Buttons   []string
+	Selected  int
+	Borders   *term.BorderSet
+	OnButton  []func()
+	OnDismiss func()
+	btnHits   []HitRegion
 }
 
 func NewConfirmDialogWidget(message string) *ConfirmDialogWidget {
@@ -54,15 +56,16 @@ func (d *ConfirmDialogWidget) Render(surface *RenderSurface) {
 	for _, btn := range d.Buttons {
 		btnW += len([]rune(btn)) + 4
 	}
-	boxW := 30
-	if msgW > boxW {
-		boxW = msgW
-	}
+	boxW := msgW
 	if btnW > boxW {
 		boxW = btnW
 	}
-	if boxW > sw-4 {
-		boxW = sw - 4
+	maxW := 60
+	if sw-4 < maxW {
+		maxW = sw - 4
+	}
+	if boxW > maxW {
+		boxW = maxW
 	}
 	boxX := (sw - boxW) / 2
 	boxY := 2
@@ -157,7 +160,8 @@ func (d *ConfirmDialogWidget) HandleEvent(ev tcell.Event) EventResult {
 		return EventConsumed
 	case tcell.KeyRune:
 		for i, btn := range d.Buttons {
-			if len(btn) > 0 && (kev.Rune() == rune(btn[0]) || kev.Rune() == rune(btn[0]+32)) {
+			first := []rune(btn)
+			if len(first) > 0 && unicode.ToLower(kev.Rune()) == unicode.ToLower(first[0]) {
 				d.Selected = i
 				if d.OnButton[i] != nil {
 					d.OnButton[i]()
--- a/internal/ui/confirmdialog_widget_test.go
+++ b/internal/ui/confirmdialog_widget_test.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"testing"
 
+	"github.com/eugenioenko/ttt/internal/term"
 	"github.com/gdamore/tcell/v2"
 )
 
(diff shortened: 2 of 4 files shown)


eugenioenko:
The auto-sizing logic and hit region refactoring significantly improve the robustness and look of dialogs. One minor concern:

The hardcoded maxW = 60 columns in ConfirmDialogWidget.Render might be too aggressive for very long error messages or dialogs with many buttons. If the content exceeds this width, it may bleed out of the border or be truncated. Consider making this limit more flexible or ensure the text rendering handles wrapping/truncation gracefully within the box.

PR #153, opened 2026-06-14
feat: distinct autocomplete completion kind symbols

eugenioenko:
## Summary
- Each LSP completion kind now shows a unique character in the autocomplete popup instead of the generic `■` square
- Added 7 new `CompletionKind` constants: Interface, Enum, EnumMember, Text, Value, File, Folder
- Updated `lspKindToUI()` to map LSP kinds to the new finer-grained UI kinds (e.g., Interface and Enum are no longer collapsed into Type)

Character mapping: `f`=function, `m`=method, `v`=variable, `c`=constant, `C`=class/type, `p`=field/property, `k`=keyword, `S`=snippet, `M`=module, `I`=interface, `E`=enum, `e`=enum member, `t`=text, `V`=value, `F`=file, `D`=folder, `■`=unknown/other.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes (all unit + E2E tests)
- Visual-only change — no new tests needed

### DIFF
--- a/internal/app/lsp_convert.go
+++ b/internal/app/lsp_convert.go
@@ -111,11 +111,13 @@ func lspKindToUI(kind lsp.CompletionItemKind) ui.CompletionKind {
 		return ui.CompletionMethod
 	case lsp.CIKVariable:
 		return ui.CompletionVariable
-	case lsp.CIKConstant, lsp.CIKEnumMember:
+	case lsp.CIKConstant:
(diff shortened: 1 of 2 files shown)


eugenioenko:
the little square with semantic colors is enough. the letters doesn't make it any less obvious the abbreviation is not helping

PR #172, opened 2026-06-14
fix: cache bracket color map to avoid per-frame recomputation

### LINKED ISSUE #179: refactor: move maxLineWidthDirty invalidation to FlushOnChange
## Summary

`maxLineWidthDirty` is currently set in `exec()` alongside `bufferDirty`. Following the same pattern we applied to `bracketColorDirty` in #172, it should move into `FlushOnChange()` so that `exec()` only sets `bufferDirty = true` and all derived cache invalidations happen in one place.

## Motivation

Keeps `exec()` minimal — it sets one flag, and `FlushOnChange()` is the single point that reacts to buffer mutations. Makes it easier to add future caches without touching `exec()`.

## Details

- Remove `e.maxLineWidthDirty = true` from `exec()`
- Add `e.maxLineWidthDirty = true` inside `FlushOnChange()` (next to `bracketColorDirty`)
- Verify any other call sites that set `maxLineWidthDirty` directly still make sense

eugenioenko:
## Summary
- Cache the `bracketColorMap` result from `computeBracketColors()` and only recompute when the buffer changes, instead of on every render frame
- Previously, scrolling a 10,000-line file to the bottom caused 10,000 lines to be scanned on every render frame; now it only scans once per edit
- Simplified `computeBracketColors()` to scan the entire file (since it's cached) and removed the `visibleStart`/`visibleEnd` parameter complexity
- Centralized both `bracketColorDirty` and `maxLineWidthDirty` invalidation into `FlushOnChange()`, so `exec()` only sets `bufferDirty` and all derived cache invalidations happen in one place

## Changes
- Added `bracketColorCache` and `bracketColorDirty` fields to `EditorPaneWidget`
- `bracketColorDirty` and `maxLineWidthDirty` are set in `FlushOnChange()` when `bufferDirty` is true — no longer scattered across `exec()` and multi-cursor methods
- Explicit `InvalidateBracketColors()` / `InvalidateMaxLineWidth()` remain for non-edit cases (tab switch, feature toggle)
- `computeBracketColors()` now takes no parameters and computes colors for all lines in the file

Closes #179

## Test plan
- [x] `make build` succeeds
- [x] All existing bracket colorization E2E tests pass (`TestBracketPairColorization`, `TestBracketPairColorizationToggle`, `TestBracketColorizationSkipsStrings`)
- [x] All other E2E and unit tests pass
- [ ] Manual verification: open a large file, scroll to bottom, verify bracket colors are correct and scrolling is smooth

### DIFF
--- a/internal/app/commands_options.go
+++ b/internal/app/commands_options.go
@@ -25,6 +25,7 @@ func (a *App) ToggleBracketPairColorization() {
 	a.EditorGroup.BracketPairColorization = a.Settings.Editor.BracketPairColorization
 	if a.EditorGroup.Editor != nil {
 		a.EditorGroup.Editor.BracketPairColorization = a.Settings.Editor.BracketPairColorization
+		a.EditorGroup.Editor.InvalidateBracketColors()
 	}
 	config.SaveSettings(*a.Settings)
 }
--- a/internal/ui/editor_group.go
+++ b/internal/ui/editor_group.go
@@ -959,6 +959,7 @@ func (g *EditorGroupWidget) syncTabs() {
 		g.Editor.LineChanges = t.LineChanges
 		g.Editor.buildDiagIndex()
 		g.Editor.InvalidateMaxLineWidth()
+		g.Editor.InvalidateBracketColors()
 		if t.TabSize > 0 {
 			g.Editor.TabSize = t.TabSize
 		}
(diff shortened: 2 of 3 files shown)


eugenioenko:
Superseded by #181 which includes the caching from this PR plus additional optimizations (dropping chroma tokenization, slices instead of nested maps).

PR #218, opened 2026-06-20
test: playground code for PR comment testing

eugenioenko:
## Summary
Sample code with deliberate review opportunities for testing PR comment features.

**Do not merge** — this is a test PR for exercising comment UIs.

### DIFF
--- a/internal/example/convert.go
+++ b/internal/example/convert.go
@@ -1,0 +1,54 @@
+package example
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+func IntToString(n int) string {
+	return strconv.Itoa(n)
+}
+
+func StringToInt(s string) (int, error) {
+	return strconv.Atoi(s)
+}
+
+func ToBinary(n int) string {
+	return strconv.FormatInt(int64(n), 2)
+}
+
+func ToHex(n int) string {
+	return fmt.Sprintf("%x", n)
+}
+
+func CelsiusToFahrenheit(c float64) float64 {
+	return c*9/5 + 32
+}
+
+func FahrenheitToCelsius(f float64) float64 {
+	return (f - 32) * 5 / 9
+}
+
+// TODO: handle negative bytes
+func BytesToHuman(bytes int64) string {
+	units := []string{"B", "KB", "MB", "GB", "TB"}
+	size := float64(bytes)
+	unit := 0
+	for size >= 1024 && unit < len(units)-1 {
+		size /= 1024
+		unit++
+	}
+	return fmt.Sprintf("%.1f %s", size, units[unit])
+}
+
+func SliceToCSV(items []string) string {
+	return strings.Join(items, ",")
+}
+
+func CSVToSlice(csv string) []string {
+	if csv == "" {
+		return nil
+	}
+	return strings.Split(csv, ",")
+}
--- a/internal/example/math.go
+++ b/internal/example/math.go
@@ -1,0 +1,67 @@
+package example
+
+import "math"
+
+func Abs(n int) int {
+	if n < 0 {
+		return -n
+	}
+	return n
+}
+
+func Max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func Min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func Clamp(val, lo, hi int) int {
+	if val < lo {
+		return lo
+	}
+	if val > hi {
+		return hi
+	}
+	return val
+}
+
+func Distance(x1, y1, x2, y2 float64) float64 {
+	dx := x2 - x1
+	dy := y2 - y1
+	return math.Sqrt(dx*dx + dy*dy)
+}
+
+// FIXME: overflows for n > 20 on 64-bit, no guard
+func Factorial(n int) int {
+	if n <= 1 {
+		return 1
+	}
+	result := 1
+	for i := 2; i <= n; i++ {
+		result *= i
+	}
+	return result
+}
+
+// this shadows builtin max, probably bad
+func max2(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func GCD(a, b int) int {
+	for b != 0 {
+		a, b = b, a%b
+	}
+	return a
+}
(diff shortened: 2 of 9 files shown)


eugenioenko:
Overall this looks like a good start, but there are several issues that need addressing before merge. See my inline comments below.

eugenioenko:
I'd also suggest adding table-driven tests for the math functions. The current test coverage is thin.

eugenioenko:
One more thing: the `example` package name is too generic. Consider renaming to something more descriptive like `playground` or `sandbox`.

internal/example/playground.go
```diff
@@ -0,0 +1,42 @@
...
+func Multiply(a, b int) int {
+	result := 0
+	for i := 0; i < b; i++ {
+		result += a
+	}
+	return result
+}
+
+func Divide(a, b int) int {
+	return a / b
```
eugenioenko, line 22: Bug: this will panic on division by zero. Add a guard: `if b == 0 { return 0 }` or return an error.

internal/example/playground.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func Fibonacci(n int) int {
+	if n <= 1 {
+		return n
+	}
```
eugenioenko, line 28: Performance: this naive recursive Fibonacci is O(2^n). Use memoization or an iterative approach. This will be unusably slow for n > 30.

internal/example/playground.go
```diff
@@ -0,0 +1,42 @@
...
+func Greet(name string) string {
+	return fmt.Sprintf("Hello, %s!", name)
+}
+
+func Add(a, b int) int {
+	return a + b
+}
+
+func Multiply(a, b int) int {
+	result := 0
```
eugenioenko, line 14: Nit: why use a loop here instead of just `return a * b`? This also breaks for negative `b`.

internal/example/validate.go
```diff
@@ -0,0 +1,61 @@
+package example
+
+import (
+	"regexp"
+	"strings"
+)
+
+// TODO: this compiles the regex every call, should be a package-level var
+func IsEmail(s string) bool {
+	re := regexp.MustCompile(`^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`)
```
eugenioenko, line 10: Performance: `regexp.MustCompile` is called every invocation. Move this to a package-level `var emailRe = regexp.MustCompile(...)`.

internal/example/math.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	result := 1
+	for i := 2; i <= n; i++ {
+		result *= i
+	}
+	return result
+}
+
+// this shadows builtin max, probably bad
+func max2(a, b int) int {
```
eugenioenko, line 55: This shadows the builtin max (Go 1.21+). Either use the builtin directly or pick a different name. Also this function appears unused.

internal/example/strings.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// REVIEW: doesn't handle consecutive capitals like "HTTPServer" -> "h_t_t_p_server"
+func CamelToSnake(s string) string {
+	var result strings.Builder
+	for i, r := range s {
+		if r >= 'A' && r <= 'Z' {
+			if i > 0 {
+				result.WriteByte('_')
+			}
```
eugenioenko, line 37: The REVIEW comment nails it: HTTPServer becomes h_t_t_p_server. Consider tracking runs of uppercase characters.

internal/example/convert.go
```diff
@@ -0,0 +1,54 @@
...
+func ToHex(n int) string {
+	return fmt.Sprintf("%x", n)
+}
+
+func CelsiusToFahrenheit(c float64) float64 {
+	return c*9/5 + 32
+}
+
+func FahrenheitToCelsius(f float64) float64 {
+	return (f - 32) * 5 / 9
```
eugenioenko, line 30: Negative bytes will produce confusing output. Add a guard or use Abs.

internal/example/playground.go
```diff
@@ -0,0 +1,42 @@
+package example
+
+import "fmt"
+
+func Greet(name string) string {
```
eugenioenko, line 5: Nit: exported functions should have godoc comments.

internal/example/strings.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return string(runes)
+}
+
+func IsPalindrome(s string) bool {
+	s = strings.ToLower(s)
+	return s == Reverse(s)
+}
+
+func CountWords(s string) int {
```
eugenioenko, line 18: Nit: CountWords could be WordCount to match Go's noun-first convention (like len, cap).

internal/example/strings.go
```diff
@@ -0,0 +1,44 @@
...
+	words := strings.Fields(s)
+	return len(words)
+}
+
+func Truncate(s string, maxLen int) string {
+	if len(s) <= maxLen {
```
eugenioenko, line 24: Nit: len(s) counts bytes, not runes. Truncating at a byte boundary could split a multi-byte character. Use []rune(s) for correctness.

internal/example/stack.go
```diff
@@ -0,0 +1,37 @@
+package example
+
+type Stack struct {
```
eugenioenko, line 3: Nit: consider using generics (Stack[T any]) instead of any for type safety.

internal/example/playground_test.go
```diff
@@ -0,0 +1,48 @@
+package example
```
eugenioenko, line 1: Nit: only playground.go has tests. The other 6 files have zero test coverage.

internal/example/convert.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return fmt.Sprintf("%.1f %s", size, units[unit])
+}
+
+func SliceToCSV(items []string) string {
+	return strings.Join(items, ",")
+}
+
+func CSVToSlice(csv string) []string {
+	if csv == "" {
```
eugenioenko, line 50: Nit: this does not trim whitespace. "a, b, c" returns [" b", " c"]. Consider strings.TrimSpace on each element.

internal/example/playground_test.go
eugenioenko: Testing

PR #273, opened 2026-06-27
fix: update placeholder vendorHash

pirate-boop:


eugenioenko (approved):
@pirate-boop  This looks great! 
Thank you for taking the time and updating to the correct hash, really appreciate it

PR #336, opened 2026-07-05
Update vendorHash for ttt buildGoModule

pirate-boop:


eugenioenko:
This update hash is for the current main right?

eugenioenko:
Thank you! I'll follow up in the issue and we'll add the CI workflow for this :)

Issue #3, opened 2026-05-26
List


eugenioenko:
- show hidden files in Files explorer

eugenioenko:
highlight active editor file in explorer

eugenioenko:
copy from diff past into search not workig ghostty macos.

eugenioenko:
typing pair of "" or () needs to be removed

eugenioenko:
last line always shows git diff green

eugenioenko:
lsp hover triggers all the time but also it doesnt dismiss after moving cursor. 
an option can be made to disable lsp hover

eugenioenko:
copy github permalink action

PR #341, opened 2026-07-06
feat: add spell checking via aspell

tenox7:
Adds native spell checking backed by `aspell`, modeled on micro's aspell plugin but integrated with ttt's existing diagnostics infrastructure.

## What it does

- Checks **prose filetypes only** — markdown, HTML, XML, TeX, reStructuredText, Org, plain text (files with no lexer, e.g. `COMMIT_EDITMSG`). Code files are skipped. Aspell filter modes (`--mode=markdown` etc.) keep code spans and URLs inside markdown out of the results.
- Misspelled words get **curly underlines** via the same `UlStyle` cell channel as LSP diagnostics. Diagnostics take precedence when both hit a cell. New theme style `editor.spellError` (defaults to the warning color).
- **Right-click a misspelled word** → up to 5 `Correct to "…"` items plus `Add "…" to Dictionary` prepended to the editor context menu.
- **Command palette**: `Spell: Suggest Corrections` (autocomplete popup at the cursor), `Spell: Add Word to Dictionary` (aspell personal dictionary), `Toggle Spell Check`.
- **Options menu** gets a Spell Check toggle with live apply (no restart). Off by default; requires `aspell` in PATH (warns when missing, inert otherwise).
- Settings: `spell.enabled`, `spell.lang` (aspell dictionary code, empty = locale), `spell.debounce` (default 500ms).

## How it works

- `internal/spell` — small aspell client speaking the ispell pipe protocol: terse mode, `^`-prefixed data lines, parses `&`/`#` result lines. Offsets are rune positions (verified with multibyte text), matching the editor's rune-based column convention.
- Async flow mirrors the git gutter: `ScheduleSpellCheck` (debounced on change) → `RequestSpellCheck` (generation counter + buffer-line copy + goroutine) → `SpellResult` posted via `EventInterrupt`. Re-checks on edit, file open, and tab switch; results are stored per tab and restored on switch, like diagnostics.
- Corrections apply as a **single-undo** `BatchCommand` (delete + paste), and the span is re-verified against the buffer first so a stale result can never mangle text — if the word moved, it just re-checks.
- Errors from aspell (e.g. missing dictionary for `spell.lang`) surface once in the status bar instead of spamming on every keystroke.

## Also included

- `rclick X Y` command in the `--exec` debug harness (needed to test the context-menu path; generally useful).
- Debug state dump now includes the active editor's misspellings.
- Regenerated reference `config/settings.json` (new `spell` section).

## Testing

- **Unit** (`internal/spell`): protocol parser (banner, blank-line tracking, `&`/`#`, suggestion cap, malformed lines), mode mapping, plus a real-aspell round-trip with multibyte offsets (skips when aspell is absent).
- **E2E** (`tests/e2e/spellcheck_test.go`): async check marks spans and sets `UlStyle` on the right cells (and not on correct words), suggest→Enter applies a correction and a single undo restores it, code files are skipped and stale spans cleared, toggle-off clears spans. All skip when aspell is missing.
- **Functional** (`tests/functional/spell-check.test.js`): toggle → suggest → accept flow and the no-misspelling notice, gated on aspell presence.
- Verified live via the debug harness: right-click menu with corrections, click-to-apply, settings persistence across runs.

## Known limitations

- Spell checking is whole-file per check (like micro's plugin); fine at typical prose sizes, guarded by a 15s timeout.
- A spell underline and an LSP diagnostic on the same cell can't stack — diagnostics win.
- Suggestions come from a completed check, so a word typed &lt;debounce ago may not offer corrections yet.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,6 +47,8 @@ The codebase follows a strict layered architecture: **core → view → render �
 
 - **`internal/workspace/`** — Multi-folder workspace management. `Folder` and `Workspace` types track one or more project roots, with `IsRepo` git-detection, `FolderForFile` lookup (longest-prefix match), and JSON-based workspace file loading/saving (`.ttt` files). The editor falls back to `cwd` when no folders are explicitly provided.
 
+- **`internal/spell/`** — Spell checking via the `aspell` binary (ispell pipe protocol). Prose filetypes only (`ModeForLanguage` maps highlighter language → aspell filter mode; code files are skipped). Misspellings render as curly underlines through the same `UlStyle` cell channel as LSP diagnostics (diagnostics take precedence). Async flow mirrors the git gutter: `ScheduleSpellCheck` (debounced, `spell.debounce`) → `RequestSpellCheck` (generation counter, line copy, goroutine) → `SpellResult` via `EventInterrupt`. Corrections apply as a single-undo `BatchCommand` after verifying the span is not stale. Off by default (`spell.enabled`); toggled from the Options menu. Suggestions surface via `spell.suggest` (autocomplete popup) and the editor right-click menu.
+
 - **`cmd/ttt/main.go`** — Entry point with event loop. Wires all components together, handles key dispatch, viewport scrolling, and redraw. Accepts a `--workspace <file>` flag to open a saved workspace, or folder/file paths as positional arguments.
 
 ### Design Principles
@@ -174,6 +176,7 @@ cat /tmp/state.json   # see full widget tree, focus, selection, panels
 
 Supported commands:
 - `click X Y` — simulate mouse click at coordinates
+- `rclick X Y` — simulate right click at coordinates
 - `hover X Y` — simulate mouse hover (move) at coordinates
 - `key COMBO` — simulate key press (e.g. `key ctrl+p`, `key enter`, `key ctrl+k x`)
 - `type TEXT` — type a string of text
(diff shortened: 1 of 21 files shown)


eugenioenko (commented):
Hi @tenox7 , thank you for putting this feature together.  It's looking awesome! 

Spell checker is actually one of the few plugins I used to install in vscode. So this was genuinly missing feature.

 On the implementation: this PR bakes the whole spell-check feature into the binary. Shelling out to an external tool like aspell is totally fine (there's precedent with ripgrep) but I'd really like to steer this class of feature toward the plugin system rather than core, so the editor stays lean and the community can own and iterate on it. I've just landed the missing pieces to make that possible: a diagnostics API (squiggles + the Diagnostics panel), an editor context-menu API, and a byte and rune column helper.



Would you be able to re-create this as a plugin please? 



It can be published as its own repo. And then you can open a PR here to update `community-plugins.json` with the link to your spellcheck plugin.


Here some info for reference that might help sort out the dirrection:

Demo PR with the functionality of spell checking using the build in diagnostic system (squigly error highlight and context menu)
https://github.com/eugenioenko/ttt-plugins/pull/3

A functioning docker plugin (example of how to interract with external binaries):
https://github.com/eugenioenko/ttt-plugins/blob/main/docker-manager/init.lua

Plugin authoring docs:
https://tttedit.dev/guides/plugin-authoring/

To quickly iterate on plugins, you can have a `/plugins` folder in the repo root and ttt will load them from there as well. BTW, plugins can be published on any repository so you can have a repo of your own with ttt-plugins and make them findeable in the binary by adding to community-plugins.json here 

Once again, thank you so much for contributing to TTT and making it an amazing terminal IDE

PR #342, opened 2026-07-06
feat: plugin diagnostics + editor context-menu API

eugenioenko:
Lets Lua plugins publish editor **diagnostics** (curly-underline squiggles) and contribute items to the editor's **right-click menu** — so linters/checkers (spell check, coverage, review comments, …) can be built as plugins instead of baked into core. It reuses the existing diagnostics rendering pipeline rather than duplicating it.

Proven end-to-end with a throwaway spell-check plugin (not included): squiggles with a custom colour, right-click "Correct to '…'", single-undo apply, and re-scan.

## Plugin API

**`ttt.diagnostics`** (new `editor.diagnostics` permission)
```lua
local diag = require("ttt.diagnostics")
diag.publish(path, {
  { line=, col=, end_line=, end_col=,        -- 1-based rune columns
    severity="error"|"warning"|"info"|"hint",
    style=<named style>,                       -- optional; overrides squiggle colour
    message=, source= },
})
diag.clear(path)   -- or diag.clear() for all this plugin's paths
```

**`editor.register_context_menu`** (gated by `editor.read`)
```lua
editor.register_context_menu(function(line, col, word)  -- 1-based, word under click
  return {
    { label = "Correct to 'and'", on_select = function() ... end },
    { separator = true },
    { label = "Add to dictionary", on_select = function() ... end },
  }
end)
```

## Core wiring

- **Layered diagnostics**: `EditorGroup` now stores diagnostics per *source* (`lsp`, `plugin:<name>`); `SetDiagnostics` routes through `SetDiagnosticsSource`, so LSP and plugin diagnostics coexist and merge. A plugin's diagnostics are cleared on uninstall/reload.
- **Custom squiggle colours**: `Diagnostic` gains an optional `Style`; `diagStyleAt` honours it. The renderer now forces a **curly** underline coloured by the style's foreground when the style defines no underline of its own — so any named colour style renders as a visible squiggle. LSP styles already define curly and are unaffected.
- **`tab.change` event**: fires when the active file changes (CLI-opened files, tab switches). Enabling a plugin mid-session also dispatches it, so a linter scans the already-open file immediately.

## Observability (generally useful)

- `rclick X Y` in the `--exec` debug harness (opens context menus).
- Active-editor diagnostics included in the debug state dump (a text screenshot can't show underlines).

## Tests

Unit tests for the diagnostics binding, the context-menu provider, the per-source diagnostic merge, and the curly-underline render fallback. `go build ./...`, the four touched packages' tests, `gofmt`, and `go vet` are all clean.

## Follow-ups (not in this PR)

- `ttt.notify(msg, level)` — status-bar messages (needed for a linter's "binary missing" hint to match core UX).
- A rune-offset helper (`byte_to_col`/`col_to_byte`) — Lua's byte offsets vs. the editor's rune columns are a footgun for text-position plugins.
- Interactive/sticky hover with code-actions (quick-fix), and Options-menu contribution.
- docs-web section for the new APIs.


eugenioenko:
Follow-up (c9d264f): dropped the enable-time `tab.change` dispatch per review. It was a global dispatch (re-scanned every plugin), duplicated across two handlers, and semantically wrong (the tab didn't change). `tab.change` now fires from a single place — the event loop, on an actual active-file change. A plugin enabled mid-session applies on the next file switch or on restart (reload-to-apply), rather than being force-scanned immediately.

PR #353, opened 2026-07-06
Add workflow to auto-update Flake vendorHash

pirate-boop:
This workflow automatically updates the vendorHash in flake.nix when changes are detected in go.mod or go.sum. It creates a pull request if the hash is updated.


eugenioenko:
🎉 

This is awesome!  @pirate-boop thank you so much for putting this together 🥇 
Really appreciate the effort! 

We can test it later by bumping go.mod deps versions which is probably way overdue now :)

PR #422, opened 2026-07-27
Update vendorHash for ttt in flake.nix

pirate-boop:


eugenioenko (approved):
Thank you for the fix! 

I wonder why the workflow didn't get triggered?

I did update go mod files in this PR, that should have triggered it
https://github.com/eugenioenko/ttt/pull/419

eugenioenko:
Update, it did run!

https://github.com/eugenioenko/ttt/actions/runs/29977981023

It doesn't have the right permissions, I'll update those later. In the meantime, thank you for the manual fix!

`1.2.0` hasn't been released yet, but it will come with tcell v3 which improves performance somewhat :)

pirate-boop:
No problem, glad to help! Looking forward to v1.2.0 and the tcell v3 performance boost

PR #425, opened 2026-07-31
fix(ui): check force keys only while a raw key consumer has focus

eugenioenko:
## Problem

`ForceKeyCommands` is documented in `internal/config/keybindings.go` as:

> ForceKeyCommands are checked even when a raw key consumer (e.g. terminal) has focus.

But `Root.HandleEvent` checked them **unconditionally**, as the very first thing on every key event — much broader than that contract. The `RawKeyConsumer` check itself lives far below, at `root.go:201`.

The effect is that five keys are permanently unreachable by a plugin key interceptor:

| key | command | Emacs meaning |
|---|---|---|
| `ctrl+b` | `sidebar.toggle` | backward-char |
| `ctrl+p` | `command.palette` | previous-line |
| `ctrl+t` | `terminal.toggle` | transpose-chars |
| `ctrl+q` | `editor.quit` | quoted-insert |
| `alt+t` | `terminal.fullscreen` | transpose-words |

This blocks an Emacs keybinding layer outright — three of those are among the keys an Emacs user presses most. `ctrl+t` was the worst of them: it moved focus to the terminal panel, and the interceptor only runs while the editor group has focus, so a single keystroke took the whole plugin offline until focus returned.

It also explains a latent bug in ttt-vim, which declares `ctrl-b` as page-up today — that binding has always been dead code.

## Change

Gate the force-key pass on a raw key consumer actually having focus, which is the case the escape hatch exists for. Extracted as `rawKeysFocused()` and reused at the existing call site.

## Why this is backwards compatible

`BindKeys` (`internal/app/commands.go:55`) calls `AddGlobalKey` **unconditionally**; the `AddForceKey` on the next line is an *additional* registration. So with no interceptor installed, these bindings are still reached through `handleGlobalKeys` and behave identically.

The only observable change is for a plugin that installs a `KeyInterceptor` — which today means ttt-vim, and there the sole overlap is `ctrl+b` becoming page-up, i.e. what the plugin always intended.

The invariant force keys exist for is preserved exactly: while the terminal is consuming raw keys they are still checked first, so a plugin can never trap the keyboard.

## Tests

Four e2e tests in `tests/e2e/key_interceptor_test.go`:

- interceptor receives `ctrl+t` when the editor has focus, and preempts the force key
- force key preempts the interceptor while a raw key consumer has focus (the invariant)
- force-key binding still fires with no interceptor installed (backwards compatibility)
- force-key binding still fires when the interceptor declines

Plus manual verification with a probe plugin claiming `ctrl+b`/`ctrl+p`/`ctrl+t` on **stock keybindings**:

| | reached plugin | terminal opened |
|---|---|---|
| before | none | yes |
| after | all three | no |

And with no plugin loaded, `ctrl+t` still toggles the terminal.

`make test` is green.

https://claude.ai/code/session_011V2yCew6ocjCHAoTwKuSqQ

### DIFF
--- a/internal/ui/root.go
+++ b/internal/ui/root.go
@@ -148,10 +148,12 @@ func (r *Root) rawKeysFocused() bool {
 
 func (r *Root) HandleEvent(ev tcell.Event) EventResult {
 	kev, isKey := ev.(*tcell.EventKey)
-	// ForceKeys are the escape hatch out of a raw key consumer, so they only
-	// outrank everything while one has focus. Otherwise they are reached below
-	// through handleGlobalKeys like any other binding.
-	if isKey && r.rawKeysFocused() {
+	// ForceKeys punch through two things that would otherwise swallow them: a
+	// raw key consumer, and an overlay (a second ctrl+q force-quits past the
+	// unsaved-changes dialog). With neither present nothing is swallowing keys,
+	// so they are left to handleGlobalKeys below like any other binding --
+	// which is what lets a key interceptor claim them while editing.
+	if isKey && (r.rawKeysFocused() || r.HasOverlay()) {
 		for _, gk := range r.ForceKeys {
 			if matchKey(kev, gk) {
 				gk.Handler()
--- a/internal/ui/root.go
+++ b/internal/ui/root.go
@@ -141,9 +141,24 @@ func matchKeyChord(kev *tcell.EventKey, gk GlobalKeyBinding) bool {
 		kev.Modifiers() == gk.Mod
 }
 
+// rawKeysFocused reports whether the focused widget is currently swallowing
+// every key event (the integrated terminal). A nil or non-consumer Focused
+// fails the type assertion and returns false.
+func (r *Root) rawKeysFocused() bool {
+	rk, ok := r.Focused.(RawKeyConsumer)
+	return ok && rk.WantsRawKeys()
+}
+
 func (r *Root) HandleEvent(ev tcell.Event) EventResult {
 	kev, isKey := ev.(*tcell.EventKey)
-	if isKey {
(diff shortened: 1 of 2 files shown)


eugenioenko:
Good catch — the original gate was wrong, now fixed in `9dd7d16`.

**What I got wrong.** I claimed force keys exist for one reason (escaping a raw key consumer) and that gating on that changed nothing else. They actually punch through **two** things: a raw consumer *and* an overlay. The overlay bypass is load-bearing — a second `ctrl+q` force-quits past the unsaved-changes dialog, which is exactly what `tests/functional/quit-confirm.test.js > should force quit with second Ctrl+Q` covers.

**Fix:** gate on either condition.

```go
if isKey && (r.rawKeysFocused() || r.HasOverlay()) {
```

With neither present nothing is swallowing keys, so they still reach `handleGlobalKeys` — which is what lets a key interceptor claim them while editing. Both original purposes are preserved.

**How it slipped through:** I verified with `make test`, which is `go test ./...` and does not run `tests/functional/`. That's on me. Now verified with both:

- `make test` — green
- `cd tests/functional && npx vitest run` — 232 passed, 27 expected fail (back to baseline; was 231/1 failed)
- e2e regression test added for the overlay case, so this can't regress silently again

Re-verified the original goal still holds: with stock keybindings a probe plugin claiming `ctrl+b`/`ctrl+p`/`ctrl+t` receives all three, and the terminal stays closed.

**On making it a setting** — I'd still lean against it. The failure was a missing condition in the gate, not evidence that the behaviour needs to be user-tunable; with the overlay case restored there's no remaining scenario I can find where the old and new behaviour differ for someone without a key-intercepting plugin. Happy to add `keybindings.forceKeys` as a follow-up if you'd rather have the escape hatch, but it'd be speculative right now.

PR #439, opened 2026-08-05
fix: highlight multiline block comments by tokenizing full buffer

eugenioenko:
## Problem

Chroma's `HighlightLine` tokenized each line in isolation with a fresh lexer state, so the JS block-comment regex (`/\*.*?\*/` with `dot_all`) could never match across line boundaries — the opener and closer were never in the same input.

- `// this comment` — worked ✓
- `/* inline comment */` — worked ✓
- `/*` / `   multiline` / `*/` — **broken** ✗

## Fix

Added `HighlightAll(lines []string) [][]Span` which joins all lines and tokenizes the full buffer in one chroma call, then splits the resulting tokens back into per-line spans. The result is cached and invalidated on `ClearCache()` (which the editor already calls on buffer changes), so there's no per-frame cost — re-tokenization only happens when the buffer actually changes.

## Changes

- `internal/core/highlight/highlighter.go` — added `HighlightAll` method with caching
- `internal/ui/editor_widget_render.go` — precomputes all spans before the render loop
- `internal/core/highlight/highlighter_test.go` — tests for multiline comments and cache reuse

### DIFF
--- a/internal/core/highlight/highlighter.go
+++ b/internal/core/highlight/highlighter.go
@@ -15,8 +15,9 @@ type Span struct {
 }
 
 type Highlighter struct {
-	lexer chroma.Lexer
-	cache map[string][]Span
+	lexer     chroma.Lexer
+	cache     map[string][]Span
+	lineSpans [][]Span
 }
 
 func New(filename string) *Highlighter {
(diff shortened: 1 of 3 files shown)


eugenioenko:
This is a regression, highlight all the file all the time intead of the modified line is a heavy price to pay

PR #442, opened 2026-08-07
fix: classify symlinked directories as expandable in explorer

eugenioenko:
This is @choephix's fix, cherry-picked from their fork with authorship preserved: I didn't write it.

**The bug:** `os.ReadDir` returns lstat-based entries, so a symlink pointing at a directory reports `IsDir() == false`. The explorer classified those as files; they sorted with files, drew a file icon, and couldn't be expanded.

**The fix:** when an entry is a symlink, `os.Stat` through it to get the real type. Broken links and symlink loops return an error and fall back to being treated as files, so nothing regresses.

Includes a test (`internal/ui/dirloader_test.go`). I confirmed it fails on `main` with `symlinked directory is not expandable` and passes here.

---

@choephix, I came across your fork while looking at how people are using ttt, and found this sitting on your `mine` branch. It's a real bug I still had, so rather than reimplementing it I've opened your commit as-is, with you as the author.

Are you okay with this being merged? Happy to drop it if you'd rather I didn't. If I don't hear back I'll plan to merge in a week or two: either way the commit stays under your name.

Thanks for it, and for `remote-open`. That one surfaced a few gaps in the plugin API that I'm filing separately.

### DIFF
--- a/internal/ui/dirloader.go
+++ b/internal/ui/dirloader.go
@@ -46,14 +46,21 @@ func LoadDirEntries(dirPath string, settings config.ExplorerSettings) []DirEntry
 			continue
 		}
 
+		isDir := entry.IsDir()
+		if entry.Type()&os.ModeSymlink != 0 {
+			if info, err := os.Stat(childPath); err == nil {
+				isDir = info.IsDir()
+			}
+		}
+
 		de := DirEntry{
 			Name:       name,
 			Path:       childPath,
-			IsDir:      entry.IsDir(),
+			IsDir:      isDir,
 			GitIgnored: isIgnored,
 		}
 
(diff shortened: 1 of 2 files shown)


choephix:
Yep! Absolutely no trouble.

Sorry for not making these real actual contributions, but I asked AI to make the fix and review it, but I never actually found the time to look myself. 😅

choephix:
Thanks so much for the great work on **`ttt`** btw, it's exactly the kind of thing I was looking for and my go-to terminal text app since I found it. ★

eugenioenko:
Thank you so much for fixing the bug and contributing to make TTT better, really appreciate it.

PR #459, opened 2026-08-20
feat(cli): open files at a line and column from the command line

arimxyer:
## What

`ttt main.go:42` and `ttt main.go:42:8` open the file with the cursor on that line and column.

```sh
ttt internal/app/widgets.go:42        # line 42
ttt internal/app/widgets.go:42:8      # line 42, column 8
ttt internal/app/widgets.go:42:       # trailing colon, as pasted from `grep -n`
```

## Why

It is the form `grep -n`, compilers, linters and language servers already print, so it is what ends up on the clipboard when you are chasing an error. Other editors accept it (`code -g file:42`, `hx file:42`, `nvim +42 file`), and without it that text has to be retyped as a Go to Line.

## Compatibility

A positional argument is only read as a position when the plain path fails to resolve, so nothing that worked before changes:

1. If something exists at the path as written, it wins. A file genuinely named `report:12` still opens as itself.
2. Otherwise the argument is retried as `path:line[:col]`, and only counts as a position when the stripped path names an existing file.
3. Otherwise it stays a literal path, so `ttt notes.txt:9` still creates a buffer with that name when `notes.txt` does not exist.

At most two trailing numeric fields are consumed, so a Windows path such as `C:\src\main.go` is untouched (the drive letter is not a number). `:0` and negative numbers are treated as part of the name, since lines are 1-based.

## Implementation note

Positioning happens in two steps, which is the one part of the diff that is not obvious.

`GoToLineCol` scrolls the target into view, and that scroll is computed against the viewport height — which is zero until the first render. Calling it at startup leaves the tab mis-framed for as long as it stays open: the target line lands just above the visible region, and the horizontal offset sticks too.

So this PR splits `PlaceCursor` out of `GoToLine`. It does the clamping and fold expansion and moves the cursor, but never touches the viewport. `GoToLine` is now `PlaceCursor` plus the existing scroll, so its behaviour is unchanged.

- Each file gets a `PlaceCursor` as it opens, while its tab is still the active one.
- The file left active is framed from the event loop after the first render, which gives it the same framing as an in-editor **Go to Line**.

`internal/ui/place_cursor_test.go` pins the no-scroll invariant, because a `scrollViewport()` added to `PlaceCursor` later would look like a tidy-up and silently bring the mis-framing back.

## Known limitation

With several positioned files, only the file left active is scrolled. The others have a correct cursor but show the top of the file until the cursor next moves, because activating a tab does not scroll to its cursor.

Framing them at startup means switching to each tab and back. When I tried that, tabs visited that way came back with their viewport shifted one column right, clipping the first character of every line. I could not reproduce that with ordinary tab switching, so it appears specific to switching during startup rather than a general problem — but I did not chase it down, and I would rather not build the feature on a mechanism I do not understand. Framing one file covers the common case of a single `file:line` argument.

If you would prefer every named file framed, say so and I will work out what that shift actually is.

## Tests

- `internal/app/args_test.go` — `splitLineCol` table, and that a suffix only resolves against a file that exists, including the colon-named-file and directory cases.
- `internal/ui/place_cursor_test.go` — clamping, and that `PlaceCursor` leaves `TopLine`/`LeftCol`/`Height` alone while `GoToLine` still scrolls.
- `tests/functional/cli-file-line-col.test.js` — six cases through the real binary: line, line+column, trailing colon, a real `report:12` file, several files, and a plain path.

`go test ./...` and the full `tests/functional` suite pass. Framing was checked against an in-editor Go to Line on the same file and size: both give the same viewport.

### DIFF
--- a/cmd/ttt/main.go
+++ b/cmd/ttt/main.go
@@ -137,6 +137,7 @@ Usage: ttt [options] [files/folders/URLs...]
 
 Arguments:
   files               Open one or more files
+  file:line[:col]     Open a file with the cursor on that line (and column)
   folders             Open directories as workspace roots
   .                   Open the current directory
   PR URL              Open a GitHub pull request for review
@@ -153,6 +154,7 @@ Options:
 Examples:
   ttt                                           Open current directory
   ttt main.go utils.go                          Open specific files
+  ttt internal/app/widgets.go:42:8              Open at line 42, column 8
   ttt ~/projectA ~/projectB                     Multi-root workspace
   ttt . https://github.com/o/r/pull/123         Review a PR with repo tree
 
@@ -211,7 +213,7 @@ Docs: https://tttedit.dev
 	cmdRegistry := command.NewRegistry()
 	borders := app.BuildBorderSet(cfg.Theme.Borders)
 
-	editor, prURLs := app.BuildApp(&cfg, &borders)
+	editor, prURLs, fileTargets := app.BuildApp(&cfg, &borders)
 	editor.ApplyBorderStyle()
 	editor.Init(screen, renderer, lspManager)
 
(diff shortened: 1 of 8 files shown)


eugenioenko (approved):
Nice! 

Thank you for taking the time and adding this, really appreciate the effort put into this.

Curios, what's your workflow with grep and ttt?

PR #461, opened 2026-08-20
fix(exec): run scripted commands on the event loop so the status bar stays in sync

### LINKED ISSUE #460: Status bar goes stale after a tab switch made by an --exec script command
The status bar keeps showing the previous tab's cursor position after switching tabs. The editor state is correct; only the rendered status bar is stale, so `Ln`/`Col` describes a file you are no longer looking at.

## Reproduce

With two files where the cursor sits on a different line in each:

```sh
printf 'a\n%.0s' {1..80} > /tmp/first.txt
printf 'b\n%.0s' {1..80} > /tmp/second.txt

bin/ttt --size 120x40 \
  --exec 'wait 400; key ctrl+g; wait 250; type 20; key enter; wait 300; exec "View: Previous Tab"; wait 400; screenshot /tmp/s.txt; debug /tmp/s.json; quit' \
  /tmp/first.txt /tmp/second.txt
```

That opens both files, jumps to line 20 in `second.txt`, then switches back to `first.txt`, whose cursor is still on line 1.

**Rendered status bar** (`/tmp/s.txt`): `Ln 20, Col 1`
**Actual state** (`/tmp/s.json`): `first.txt`, cursor line 0 → `Ln 1, Col 1`

**Correction:** I originally wrote that this also happens interactively. That is wrong, and I am sorry for the noise. Having checked each path, only the scripted one is affected:

| tab switch via | rendered | actual | |
|---|---|---|---|
| `exec "View: Previous Tab"` | `Ln 20, Col 1` | `first.txt Ln 1` | stale |
| key `alt+,` | `Ln 1, Col 1` | `first.txt Ln 1` | ok |
| mouse click on the tab | `Ln 1, Col 1` | `first.txt Ln 1` | ok |

`execKey` and `execClick` post real tcell events, so they go through the event loop and get its `syncStatus()`. `execCommand` calls `Reg.Execute` straight from the script goroutine and skips it.

So this is a bug in the `--exec` harness rather than in the editor, which still matters: `tests/functional` drives the binary through `--exec`, so a test that switches tabs with `tui.exec(...)` and asserts on the status bar gets a false failure. That is exactly how I ran into it.

## Notes

Reproduced on `c7da019` (current `main`, after #458).

The content pane updates correctly on the switch — it is only the status bar segment that lags, so the two disagree on screen.

Worth knowing when reading the reproduction: a control that leaves both files' cursors on line 1 cannot detect this, because a stale reading and a correct one are then identical. The two cursors have to differ.

Happy to send a PR if you would like it fixed this way rather than another.

arimxyer:
Fixes #460.

## What was stale

`exec "Command Name"` in an `--exec` script called `Reg.Execute` directly from the script goroutine. The event loop runs `syncStatus()` after handling real key and mouse input; a command dispatched this way never reached that, so the editor state changed while the rendered status bar kept the previous values.

Switching tabs is where it shows: after `exec "View: Previous Tab"` the position segment still reads the line from the tab that *had* been active.

```
exec "View: Previous Tab"   rendered: Ln 20, Col 1   actual: first.txt Ln 1   stale
key alt+,                   rendered: Ln 1, Col 1    actual: first.txt Ln 1   ok
click <tab>                 rendered: Ln 1, Col 1    actual: first.txt Ln 1   ok
```

Only the scripted path was affected — `execKey` and `execClick` post real tcell events and go through the loop, so interactive tab switching was always correct. The issue as I first wrote it said this also happened interactively; that was wrong, and I have corrected it.

Since `syncStatus` also drives the explorer's active file, the git gutter request, the `tab.change` plugin event, blame and the outline selection, all of those were skipped too for scripted commands.

## The fix

The command is posted as an `ExecCommandRequest` interrupt and executed on the main thread, on the same path as key and mouse input, followed by the same `FlushEditorOnChange` and `syncStatus`. This also stops the script goroutine from mutating widget state off the main thread, which `CLAUDE.md` calls out as a rule.

Four lines in the event loop, and `execCommand` posting instead of executing.

## Proving the test detects it

`tests/functional/statusbar-tab-switch.test.js` was run against a build of this branch with the fix stashed, and then with it applied:

```
PRE-FIX   exit=1   AssertionError: expected '  File   Edit   Selection   View   Op…' to contain 'Ln 1, Col 1'
POST-FIX  exit=0   Tests  1 passed (1)
```

The two files deliberately sit on different lines. With both cursors on line 1 a stale reading and a correct one are identical and the test cannot fail, so the test moves `second.txt` to line 20 first.

## Suites

`go test ./...` exit 0 (25 packages), `go vet` clean, full `tests/functional` suite exit 0 (281 passed, 27 expected fail).

`clipboard-roundtrip.test.js` failed once under parallel load and passes 3/3 in isolation; it contains no `tui.exec` calls, so it is not touched by this change.

`gofmt -l` reports `internal/ui/editor_group.go`, which is unrelated to this branch — it is already flagged on `main` and I left it alone.

### DIFF
--- a/internal/app/eventloop.go
+++ b/internal/app/eventloop.go
@@ -262,6 +262,10 @@ func RunEventLoop(
 				}
 			case *GitGutterTrigger:
 				app.RequestGitGutterForActiveFile()
+			case *ExecCommandRequest:
+				app.Reg.Execute(v.ID)
+				app.FlushEditorOnChange()
+				syncStatus()
 			case *AutocompleteTrigger:
 				triggerChar := app.charBeforeCursor()
 				isTrigger := triggerChar != "" && (len(app.CompletionTriggers) == 0 || app.isCompletionTrigger(triggerChar))
--- a/internal/app/exec_script.go
+++ b/internal/app/exec_script.go
@@ -240,7 +240,18 @@ func execCommand(a *App, args string) {
 		slog.Error("exec_script: command not found", "title", title)
 		return
 	}
-	a.Reg.Execute(cmd.ID)
+	// Hand the command to the event loop instead of running it here. `key` and
+	// `click` already post real events, so they get the status bar sync that
+	// follows real input; running a command straight from the script goroutine
+	// skipped that sync and mutated widget state off the main thread.
+	a.Screen.PostEvent(tcell.NewEventInterrupt(&ExecCommandRequest{ID: cmd.ID}))
+}
+
+// ExecCommandRequest is posted as an EventInterrupt so a command named by an
+// --exec script runs on the main thread, on the same path as key and mouse
+// input.
+type ExecCommandRequest struct {
+	ID string
 }
 
 func execScreenshot(a *App, args string) {
--- a/tests/functional/statusbar-tab-switch.test.js
+++ b/tests/functional/statusbar-tab-switch.test.js
@@ -1,0 +1,41 @@
+import { describe, it, expect, afterEach } from "vitest";
+import * as tui from "./tui.js";
+import { createTempDir, createMultiLineFile, cleanupDir } from "./helpers.js";
+
+let dir;
+
+afterEach(() => {
+  tui.kill();
+  if (dir) cleanupDir(dir);
+});
+
+describe("status bar after a scripted tab switch", () => {
+  // The two files must sit on different lines: if both cursors are on line 1 a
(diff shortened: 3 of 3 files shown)


eugenioenko (approved):
@arimxyer Thank you for putting the fix together, really appreciate the  collaboration !

This is a great catch and spot on 🎉

arimxyer:
Epic! I've got another one coming too!

PR #466, opened 2026-08-21
test: add lifecycle coverage for Terminal, fix Close() deadlock

eugenioenko:
## Summary
- `internal/terminal` had zero test coverage despite owning real PTY/process lifecycle (spawn, resize, cleanup). Adds `terminal_test.go` covering: spawn + initial size, `Close()` killing the process and stopping the read loop, `Close()` idempotency, concurrent `Close()` calls, and read-loop output actually landing in the vt10x view.
- Writing the tests surfaced a real deadlock: `Close()` unconditionally blocks on `<-t.done`, but `t.done` is only closed by `readLoop()`, which only runs if `Run()` was called first. Every current call site (`internal/app/app.go`) happens to call `Run()` immediately after `New()`, so this hasn't fired in production — but it's a landmine for any future caller. Fixed by tracking whether `Run()` started the read loop and only waiting on the channel if it did.

## Test plan
- [x] `go test ./internal/terminal/... -race -v` — all 7 tests pass
- [x] `go build ./...`
- [x] `go test ./...` — full suite green

## Summary by CodeRabbit

* **Bug Fixes**
  * Closing a terminal before it starts no longer causes it to hang.
  * Repeated and concurrent close operations are handled safely.
  * Terminal shutdown now reliably stops the running process and read loop.

* **Tests**
  * Added coverage for terminal sizing, scrollback defaults, process termination, shutdown behavior, and shell output.

### DIFF
--- a/internal/terminal/terminal.go
+++ b/internal/terminal/terminal.go
@@ -25,6 +25,7 @@ type Terminal struct {
 	cmd        *pty.Cmd
 	cols, rows int
 	done       chan struct{}
+	started    bool
 	closed     bool
 	exited     bool
 	OnUpdate   func()
@@ -89,6 +90,9 @@ func defaultShell() string {
 }
 
 func (t *Terminal) Run() {
+	t.mu.Lock()
+	t.started = true
+	t.mu.Unlock()
 	go t.readLoop()
 }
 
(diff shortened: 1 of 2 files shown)


eugenioenko:
@coderabbitai review

PR #468, opened 2026-08-21
fix(terminal): size PTY from bottom panel height, not full window

### LINKED ISSUE #465: fix(terminal): terminal PTY height initialized to full window height instead of bottom panel height
### Summary
When opening the integrated terminal in `ttt`, the terminal's PTY and `vt10x` virtual buffer are initialized with rows calculated from the full window split height (`ContentSplit.H - 3`), rather than the actual bottom panel content height (`BottomH - 2`).

This causes dynamic CLI applications (e.g. `antigravity-cli`, `claude-cli`, interactive TUIs, or any streaming chat CLI) to print into off-screen rows (e.g. rows 10–37 in a 10-row panel), resulting in:
- Output appearing frozen / missing active prompt and streaming tokens.
- Chat text failing to scroll until the full virtual height is filled, at which point the visible viewport lags dozens of lines behind the actual output.
- Cross-platform differences: on Linux, window managers often emit an initial `tcell.EventResize` on window mapping which triggers `resizeTerminals(app)` and masks the bug; on macOS, no initial resize event is sent, leaving the PTY permanently desynchronized.

### Root Cause
1. **Spawn calculation:** In `internal/app/app.go` (`SpawnTerminal`), rows are calculated from `a.ContentSplit.GetRect().H - 3`, which spans the full editor + bottom panel height.
2. **Panel resize callbacks:** In `internal/app/callbacks.go` (`ContentSplit.OnResize`) and `internal/app/commands_view.go` (`ToggleTerminal`, `ToggleTerminalFullscreen`), changing `BottomH` does not call `resizeTerminals(app)` to resize existing terminal PTYs.
3. **PTY startup resize:** In `internal/terminal/terminal.go`, `pt.Resize(cols, rows)` is called after `cmd.Start()`, which can race with child process initialization on Darwin where `openpty` starts with 0x0 dimensions.

### Steps to Reproduce
1. Open `ttt` on macOS (without resizing the outer terminal window).
2. Open the terminal panel (`ctrl+t`).
3. Run an interactive agent CLI like `antigravity-cli` or `claude-cli` (or run a long streaming command).
4. Observe that as lines stream, the visible terminal area does not scroll and recent messages/prompts remain hidden below the visible viewport.

### Proposed Fix
1. **Sizing in `SpawnTerminal` (`internal/app/app.go`):** Use `a.ContentSplit.BottomH` (minus tab bar and border offset) when calculating initial `rows`.
2. **Sizing on split resize:** Invoke `resizeTerminals(app)` in `ContentSplit.OnResize` and `ToggleTerminal` / `ToggleTerminalFullscreen`.
3. **Resize PTY before `cmd.Start()`:** In `internal/terminal/terminal.go`, set initial winsize before spawning the process.

eugenioenko:
## Summary
- `SpawnTerminal` sized the PTY/vt10x buffer from the full `ContentSplit` height instead of the bottom panel's actual height, so interactive CLIs (streaming chat tools, TUIs) rendered into off-screen rows — output appeared frozen or desynced until an unrelated resize event happened to correct it.
- Resize the PTY before `cmd.Start()` so the child process doesn't read the pty's default winsize on startup.
- Consume `WheelLeft`/`WheelRight` in the terminal widget so horizontal trackpad scroll doesn't fall through and scroll the editor underneath.
- `redraw()` rendered the frame, resized terminal PTYs, *then* read the cursor position — so on any frame where a terminal actually resized, the painted content and the reported cursor position came from two different Term states. Moved the resize to the very end of `redraw()` so both always read the same state within a frame.
- Kept eager `resizeTerminals()` calls on `ToggleTerminal`, `ToggleTerminalFullscreen`, and the split drag callback (`OnResize`) — dropping them in favor of only the trailing call above added a one-redraw-cycle lag between a panel size change and the PTY picking it up, which regressed fullscreen sizing.
- **Scrollbar drag-to-bottom fix**: dragging the terminal scrollbar to the end of the track resolved the new offset against a `TotalItems` snapshot only as fresh as the widget's last `Render()`. While a CLI actively streams output, dragging to the track's end always landed short of the true live bottom (measured 13 lines short in a test). Reaching the track's end now goes live unconditionally instead of resolving against the stale snapshot.
- **Root cause of "output not using the full panel height" once scrolled**: apps like `claude-cli` set a DECSTBM scroll region to keep an input/status bar pinned at a fixed row while only the log area above it scrolls (`htop`-style tools do the same). `vt10x` already respects this internally — content outside the region is never pushed to scrollback — but ttt's terminal widget had no way to know where the region was, so its history-scrolling render path treated the whole screen as one flat scrollable sequence. Once scrolled, rows that should always show live content (the fixed bar) went blank instead. Fixed by bumping to a `vt10x` commit that exposes `View.ScrollRegion()` ([eugenioenko/vt10x#2](https://github.com/eugenioenko/vt10x/pull/2)) and reworking `Render()`/`screenToLine()` to keep rows outside `[top, bottom]` pinned to live content regardless of `scrollOffset`, while rows inside the region keep the existing scrollback+live blending (now scoped to the region's own height). Reproduced and verified via `--exec` with a synthetic DECSTBM scroll-region script before and after the fix (fixed status row went blank when scrolled without the fix, stayed correctly pinned with it) — this also explains why `btop` never showed the bug (it redraws via absolute cursor positioning, never scrolling at all, so nothing lands in scrollback) versus `claude-cli` (streams to the primary buffer inside its scroll region, so scrollback grows and the bug was reachable).

`go.mod` is pinned to a commit on the `vt10x` PR branch (not a local `replace`), so this builds cleanly — CI included — without needing that PR merged first. Once it merges, a follow-up bump to the merged commit is needed.

Fixes #465.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (all packages pass)
- [x] Added `TestSpawnTerminalSizedToBottomPanel` (e2e) — fails against the pre-fix code (34 rows instead of 13), passes with the fix
- [x] Added `TestScrollbarDragToBottomGoesLiveDespiteStaleTotalItems` (unit, real PTY via `/bin/cat`) — fails against the pre-fix code (`scrollOffset = 13` instead of 0), passes with the fix
- [x] Added `TestScrolledRenderKeepsRowsOutsideScrollRegionLive` (unit, real PTY via `/bin/cat`, synthetic DECSTBM scroll region) — fails against the pre-fix code (fixed row renders raw escape-sequence garbage instead of the pinned text), passes with the fix
- [x] `cd tests/functional && pnpm test` (287 passing, prior to the redraw-ordering and scroll-region fixes)
- [x] `go vet ./...`, `gofmt -l` clean
- [ ] Manual verification on macOS with `claude` pending — user is testing iteratively; this push targets the confirmed root cause (scroll region not respected while scrolled)

Known, separate, out-of-scope issue found while investigating: `internal/ui/terminal_widget.go` never forwards mouse clicks to the PTY as xterm mouse-report escape sequences, so apps like `btop` that expect click-to-position (e.g. a filter box) don't receive them. Confirmed unrelated to this issue. Worth a separate issue if wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal sizing for bottom panels and fullscreen layouts.
  * Terminal dimensions now update correctly when panels are resized or opened.
  * Improved terminal startup sizing and redraw behavior.
  * Fixed terminal rendering and selection within scroll regions while preserving fixed rows.
  * Improved scrollbar behavior when returning to the live view.
  * Horizontal mouse-wheel actions no longer trigger unrelated terminal interactions.

* **Tests**
  * Added regression coverage for panel sizing, scroll regions, and scrollbar navigation.


eugenioenko:
closing in favor of solution in #471

PR #470, opened 2026-08-21
feat(debug): --listen HTTP command server + raw PTY byte capture in debug dump

### LINKED ISSUE #469: feat(debug): --listen HTTP command server + raw PTY byte capture in debug dump
## Context

`ttt` has a `--exec "commands"` debug harness (`internal/app/exec_script.go`) that runs a batch of scripted commands (click/key/type/screenshot/debug/...) once at startup, then continues normally. Documented in `ExecScriptUsage()` and the "Debug harness" section of `CLAUDE.md`.

While investigating a terminal-emulation bug (integrated terminal, `internal/terminal` wrapping the `vt10x` fork) that only reproduces during real, organic interactive use with a live CLI (e.g. `claude-cli`) — not from a scripted `--exec` batch — two gaps showed up:

1. No way to send exec-style commands to an **already-running** `ttt` process interactively, to capture a repro at the exact moment it happens instead of scripting it in advance.
2. No way to see the **raw bytes** a PTY child process actually wrote, independent of how `vt10x`/ttt parsed and rendered them — ground truth for diagnosing terminal-emulation bugs, since screenshots and the existing `debug` JSON only show ttt's own interpretation.

This issue covers both. Two independent pieces — either order, both needed for the full workflow.

### Flag
Add `--listen` (boolean, no argument) to `cliFlags` in `cmd/ttt/main.go`, alongside `--exec`/`--size`/`--debug` in `parseFlags()`. When set, after the app is fully wired up (same point where `--exec` does `go app.RunExecScriptSep(...)`), start an HTTP server in a goroutine on `127.0.0.1:4242` (hardcoded for now — a `--listen-port` override can come later, don't build it speculatively).

**Bind to `127.0.0.1` only, never `0.0.0.0`** — this executes arbitrary key/type/click input against a running editor; must not be reachable off the local machine.

### Endpoint
`POST /exec` — body is a script string in the exact same format `--exec` accepts (commands separated by `;` by default). Reuse `app.RunExecScriptSep(editor, string(body), sep)` directly — do not duplicate the command switch in `exec_script.go`. Accept an optional `?sep=` query param mirroring `--exec-split-on`, defaulting to `;`.

Response: run the script **synchronously** in the handler (block until `RunExecScriptSep` returns), respond `200 OK` with plain-text `"ok"`. Empty body → `400`.

### Wiring
In `main()`, near the existing:
```go
if flags.exec != "" {
    go app.RunExecScriptSep(editor, flags.exec, flags.execSplitOn)
}
```
add:
```go
if flags.listen {
    go startListenServer(editor)
}
```
Put `startListenServer` in a new `cmd/ttt/listen.go` (or inline, whichever keeps `main.go` reasonably sized — use judgment).

### Shutdown
we can add a "shutdown" command that will execute exit

### Docs
Update `--help` output in `main.go` and `ExecScriptUsage()` (or a sibling doc block) to mention `--listen`. Update the "Debug harness" section of `CLAUDE.md`, including an example `curl` invocation.

### Capture buffer
In `internal/terminal/terminal.go`, add a bounded ring/tail buffer to `Terminal` that accumulates every byte read from the PTY, independent of how `vt10x` interprets it:

```go
const rawTailMax = 64 * 1024

type Terminal struct {
    // ...existing fields...
    rawTail []byte
}
```

In `readLoop()`, where it currently does `t.vt.Write(buf[:n])` under `t.mu.Lock()`, also append to `t.rawTail` and trim from the front if it exceeds `rawTailMax` (keep the *most recent* bytes).

Add an accessor matching the style of `ScrollbackLen()`/`Size()` (locks, returns a copy):
```go
func (t *Terminal) RawTail() []byte {
    t.mu.Lock()
    defer t.mu.Unlock()
    out := make([]byte, len(t.rawTail))
    copy(out, t.rawTail)
    return out
}
```

### Wire into `DebugState`
In `internal/app/debug_dump.go`, add to `DebugState`:
```go
Terminals []DebugTerminal `json:"terminals,omitempty"`
```
```go
// DebugTerminal reports one integrated-terminal tab's raw PTY byte tail, so
// a terminal-emulation bug can be diagnosed from what the child process
// actually sent rather than from ttt's parsed/rendered state.
type DebugTerminal struct {
    ID      string `json:"id"`
    Cols    int    `json:"cols"`
    Rows    int    `json:"rows"`
    RawTail string `json:"raw_tail"`
}
```
In `BuildDebugState()`, iterate `a.Terminals` (`[]TerminalTab`, see `internal/app/app.go`) and populate one `DebugTerminal` per tab. Cols/rows from `Term.Size()`. For `RawTail`, use `strconv.Quote(string(tt.Term.RawTail()))` — JSON-safe, human-readable escaped representation (control chars as `\x1b`, `\n`, etc.) rather than base64, far more useful for a human or an LLM to read directly.

### Testing
Follow the project's existing conventions (`CLAUDE.md` "Testing" section):
- Unit test in `internal/terminal` for `RawTail()`: write known bytes via `WriteString`, assert `RawTail()` contains them; write more than `rawTailMax` bytes, assert the buffer is capped and holds the *tail* (most recent), not the head.
- e2e test in `tests/e2e` (mirror `terminal_link_test.go`/`panel_test.go`): open a terminal (`h.exec("terminal.toggle")`), write known content, call `h.app.DumpDebugState(path)`, parse the JSON, assert `terminals[0].raw_tail` contains the expected escaped content.
- For Part 1, an e2e/integration-level test that starts `ttt --listen` as a real subprocess (see `tests/integration/` for the pattern) and posts to `POST http://127.0.0.1:4242/exec` with a simple script (e.g. `type hi; screenshot /tmp/x.txt`), asserting the screenshot file contains "hi". If a real subprocess test is disproportionate effort, a smaller unit test against `startListenServer`'s handler via `httptest.Server` covering the request/response contract is an acceptable substitute — use judgment.

### Post-implementation review
Per `CLAUDE.md`: after both parts pass tests, review for dead code, naming consistency with `exec_script.go`/`debug_dump.go`, confirm `gofmt`/`go vet`/`go build ./...` clean. Run `go test ./...` plus `cd tests/functional && pnpm test` before considering this done.

eugenioenko:
## Summary
- `--listen` starts a loopback-only (`127.0.0.1:4242`) HTTP server; `POST /exec` runs the same script format as `--exec`, synchronously, against an already-running editor — for capturing a repro at the exact moment it happens instead of scripting it in advance. `?sep=` mirrors `--exec-split-on`. Added a `shutdown` alias for `quit` for use over the endpoint.
- `Terminal` now keeps a bounded (64KB) tail of raw bytes read from the PTY, independent of `vt10x` parsing. Exposed via `RawTail()` and surfaced in the `debug` command's JSON dump as `terminals[].raw_tail` (one entry per integrated-terminal tab) — ground truth for diagnosing terminal-emulation bugs, since screenshots and existing debug state only show ttt's own interpretation.

Closes #469.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt` clean on touched files
- [x] `go test ./...` — added unit tests for `RawTail()` capture/capping (`internal/terminal`), an e2e test asserting `debug`'s JSON dump surfaces the raw tail for a live PTY terminal (`tests/e2e`), and `httptest`-backed unit tests for the `/exec` handler's request/response contract (`internal/app`)
- [x] `cd tests/functional && pnpm test` — 86 files, all passing
- [x] Manual smoke test: ran `bin/ttt --listen` under a pty (`script -qc`), curled `POST /exec` with `type hi; screenshot ...` and confirmed "hi" landed in the screenshot, verified `400` on empty body, `405` on non-POST, and `shutdown` exiting the process

## Summary by CodeRabbit

* **New Features**
  * Added a loopback-only HTTP command server, enabled with `--listen`, for controlling a running editor via `POST /exec`.
  * Added support for `shutdown` as an alternative to `quit`.
  * Added terminal diagnostics showing dimensions and recent raw terminal output.
  * Added support for custom command separators in HTTP requests.

* **Documentation**
  * Updated command-line help and usage examples for the new server, commands, and terminal diagnostics.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -190,7 +190,7 @@ Every new feature or bug fix should include tests at multiple levels:
 
 Functional tests are the highest-value tests. Use `tui.exec("Command Name")` for command palette, `tui.pressChord("ctrl+k", "x")` for keybindings, and `tui.snapshot()` to verify results.
 
-### Debug harness (`--exec`, `--plugin`, `--size`, `--debug`)
+### Debug harness (`--exec`, `--plugin`, `--size`, `--debug`, `--listen`)
 
 **USE THIS FOR DEBUGGING AND TESTING.** The editor has a built-in scripted interaction system that is faster than TUI tests and gives you direct access to internal state — reach for it before investigating UI bugs manually.
 
@@ -213,10 +213,20 @@ Supported commands:
 - `copy` — copy the current selection to the clipboard
 - `exec "Command Name"` — run a command by title (same as command palette)
 - `screenshot PATH` — save screen text to file
-- `debug PATH` — save debug state JSON (screen, cursor, buffer, focus, panels, tabs, selection, output log, full widget tree with rect/focus/props per node)
+- `debug PATH` — save debug state JSON (screen, cursor, buffer, focus, panels, tabs, selection, output log, integrated-terminal raw PTY byte tails, full widget tree with rect/focus/props per node)
 - `wait MS` — wait milliseconds
 - `panel ID` — show and focus a bottom panel by ID
-- `quit` — exit the editor
+- `quit` / `shutdown` — exit the editor
+
+**`--listen`** — Start an HTTP command server on `127.0.0.1:4242` (loopback-only — never exposed off the local machine). `POST /exec` runs the same script format as `--exec`, synchronously, against an **already-running** editor — for capturing a repro at the exact moment it happens instead of scripting it in advance:
+
+```bash
+bin/ttt --listen &
+curl -X POST --data "type hi; screenshot /tmp/screen.txt" http://127.0.0.1:4242/exec
+curl -X POST --data "shutdown" http://127.0.0.1:4242/exec
+```
+
+Pass `?sep=` to use a different command separator, mirroring `--exec-split-on`.
 
 **`--size WxH`** — Force screen dimensions for deterministic layout (e.g. `--size 120x40`). Essential for reproducible screenshots and coordinate-based click tests.
 
--- a/README.md
+++ b/README.md
@@ -742,6 +742,7 @@ TTT includes a built-in scripted interaction system designed for AI agent intera
 | Flag | Description |
 |------|-------------|
 | `--exec "commands"` | Execute semicolon-separated commands after startup |
+| `--listen` | Start an HTTP command server on `127.0.0.1:4242` (`POST /exec` accepts the same script format as `--exec`, against an already-running editor) |
 | `--plugin FILE` | Load a Lua plugin file on startup with full permissions |
 | `--size WxH` | Force screen dimensions (e.g. `120x40`) for deterministic layout |
 | `--debug` | Enable debug mode regardless of config |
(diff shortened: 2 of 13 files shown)


eugenioenko:
@coderabbitai review

PR #474, opened 2026-08-21
feat(git): add an integrated review workspace

arimxyer:
## Summary

This turns the Changes surface into a fuller Git review workspace while tightening the surrounding editor ergonomics.

- Organize staged and unstaged files as collapsible trees, with matching expand/collapse controls in Changes and Explorer.
- Browse commit history in a resizable half-panel, open complete commit details with authored date/time, and review all current changes in one scrollable document.
- Keep repository state live while the editor is open, including external file changes and Git index updates.
- Switch between split/unified and changes-only/full-file diffs, with clearer +/- gutters and an optional high-contrast treatment.
- Consolidate presentation controls into submenus and make sidebar/editor tabs draggable to reorder.
- Reconcile the bundled theme catalog/default-theme behavior and make the integration harness reproducible by pinning `tui-use` locally.

## Verification

- `make build`
- `go test ./...`
- `cd tests/functional && pnpm test` — 96 files passed; 334 passed, 26 expected failures
- `cd tests/integration && pnpm exec vitest run --no-file-parallelism` — 8 files passed; 49 tests passed
- Default parallel integration runs each reached 48/49 with a moving shared signature-help/log timeout; both affected tests pass independently, and one also passes independently on `origin/main`, isolating the failure to shared-log/PTY contention rather than feature behavior

## Implementation notes

- Core/editor state remains owned by `App`; filesystem and Git watchers publish typed events back onto the main event loop before widgets are reconciled.
- Diff syntax colors continue to layer over semantic diff backgrounds through the existing theme/style system.
- Commit pagination appends into the existing tree instead of rebuilding it, preserving selection and expansion state.

## Demo

Media captured from deterministic throwaway repositories with `termctrl` at 120x40 using the bundled Aurora theme and high-contrast diffs. Each video covers a distinct workflow; there is no stitched replay.

<table>
  <tr>
    <td width="50%"><strong>Changed-file trees</strong><br>Staged and unstaged files organized as collapsible trees</td>
    <td width="50%"><strong>Current changes document</strong><br>All current changes in one scrollable diff document</td>
  </tr>
  <tr>
    <td width="50%"><strong>Commit details and timestamps</strong><br>Commit detail view with authored date and time</td>
    <td width="50%"><strong>Grouped diff controls</strong><br>Grouped split unified and diff-context presentation controls</td>
  </tr>
  <tr>
    <td colspan="2"><strong>Full-file diff context</strong><br>Full-file diff context with compact and expanded presentation controls</td>
  </tr>
</table>

[...]

PR #480, opened 2026-08-22
feat(plugin): support checked menu entries

arimxyer:
Extracted from #474 as an independent plugin/widget foundation.

## Summary

- add optional `checked` state to plugin menu entries used by titles, dropdowns, sidebar actions, and tree/list/table node menus
- preserve three distinct backward-compatible states:
  - omitted: legacy indicator-free layout
  - `false`: reserved unchecked slot
  - `true`: checked slot
- centralize conversion through the existing context-menu renderer
- reconcile reused dropdown/title/tree/list/table widgets so dynamic redraws update entries and callbacks instead of retaining stale state
- measure menu labels and shortcuts in terminal display columns and clamp narrow/zero-width surfaces safely
- document the Lua API contract

Existing plugins that omit `checked` retain their previous menu geometry and behavior.

## Verification

- focused app/plugin/UI/widget tests, including 20 repeated runs
- affected race tests
- full Go suite with only the four independently reproduced terminal-mouse baseline timeouts excluded
- `make build`
- checked-menu functional tests: 3/3 passing
- functional composite: 86 non-clipboard files passing; clipboard file 5/5 passing in isolation
  - the parallel shared-clipboard interference was reproduced independently on current `main` and is unrelated to this diff
- fresh exact-head runtime probes for mixed sidebar rows and dynamic dropdown `true → false → omitted → true` redraw/reopen transitions
- fullwidth label/shortcut sizing, narrow-surface hit bounds, separators, keyboard/mouse indexes, and current callback replacement
- `gofmt` and `git diff --check`

The corrected head passed independent adversarial re-review. The branch contains no Git-review, diff, tree, theme, palette, tab, repository-state, integration dependency, or permission changes.

Part of #474.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,7 +114,7 @@ Lua plugins render UI in sidebar panels, bottom-panel tabs, drawers, and editor 
 | Method | Lua fields | Description |
 |---|---|---|
 | `p:label(text)` or `p:label({...})` | `text`, `style`, `badge`, `width`, borders | Static text line. `style` is a named style (see below). `border`/`border_top`/`border_bottom`/`border_left`/`border_right` draw borders. Supports box model. |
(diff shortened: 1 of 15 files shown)


arimxyer:
@coderabbitai review

eugenioenko (approved):

PR #481, opened 2026-08-22
feat(palette): add orientation help mode

arimxyer:
Extracted from #474 as an independent command-palette discoverability feature.

## Summary

- add a `?` help mode with six curated orientation topics for workspace, panels, navigation, changes, terminal, and keybinding chords
- provide topic drill-in with precise command ranking, stable ordering, and a relevance floor that avoids unrelated result floods
- show explicit zero-result guidance and direct users to `>` for the complete command registry
- preserve current routing:
  - `Ctrl+P` opens `>` command mode
  - `Ctrl+K P` opens file search
- keep ordinary command/file palettes at the existing 60-column maximum while allowing help mode to use a responsive wider layout
- recalculate geometry, borders, selection, scrolling, and mouse hit targets when switching modes or resizing
- render long and fullwidth labels with the project text-width rules

## Verification

- focused UI and E2E tests, including race runs
- command-palette functional tests: 6/6 passing
- functional composite: clipboard file 5/5 in isolation; remaining 85 files passing with 286 passes and 27 expected failures
  - the known parallel shared-clipboard interference was independently reproduced on current `main`
- broad Go suite with only the four established terminal-mouse baseline timeouts excluded
- `make build`, formatting, and `git diff --check`
- direct 200×50 mode replay: `60 command → 120 help → 60 command → 60 file → dismissed → 60 reopen → 60 Ctrl+K P`
- width thresholds around each cap, same-widget resize, narrow/degenerate surfaces, fullwidth border safety, scrolled/truncated mouse execution, ranking stability, drill-in/Escape, and registry completeness

The corrected head passed independent adversarial re-review. This seven-file branch contains no plugin checked-menu, Git-review, diff, tree, theme, tab, or repository-state work.

Part of #474.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ Changes panel in the sidebar (Ctrl+K C) with full staging workflow.
 - **Ctrl+K P** — opens quick file open (searches all files across workspace folders)
 - Type `>` in quick-open mode to switch to command mode
 - Delete the `>` in command mode to switch to file mode
+- Type `?` to browse searchable help for panels, navigation, and key chords
 - Menu shortcuts resolve dynamically from your keybindings
 
 ### Bottom Panel
--- a/docs-web/src/content/docs/getting-started/quick-start.md
+++ b/docs-web/src/content/docs/getting-started/quick-start.md
@@ -45,7 +45,7 @@ ttt https://github.com/owner/repo/pull/123  # review a PR
 
 ## Command Palette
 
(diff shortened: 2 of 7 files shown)


eugenioenko (approved):
🔥 !

This is an awesome addition. Code looks solid. 

UX wise, one tiny  improvement that can be done as follow up whenever :

- When switching to the new SelectDialog, the size of the dialog doesn't match the size of the CommandPalette and that requires  an unexpected switch of focus, by keeping it the same size users focus can stay at the same place.

PR #482, opened 2026-08-22
feat(ui): reorder tabs by dragging

arimxyer:
## Summary

- let users drag visible sidebar and editor tab headers to reorder them, with render-derived drop targets and indicators
- persist sidebar panel order while preserving late plugin panels, and provide command and menu fallbacks for keyboard-driven reordering
- preserve editor active identity, preview commitment, and pinned-tab boundaries during moves
- capture drag ownership on press across root and split boundaries, with lifecycle-safe cancellation and stationary overflow auto-scroll
- bind gestures to stable unique tab identities so removal, replacement, duplicate panel registration, resize, and hidden geometry cannot retarget a pending drag

This is the focused tab-reordering extraction from #474.

## Verification

- `make build`
- focused widget, UI, and E2E tab-reorder suites
- focused race tests repeated 10 times for capture, identity, duplicate registration, and timer lifecycle
- functional `editor-tab-reorder.test.js` and `sidebar-tab-reorder.test.js`
- full functional suite before final rebase: 88 files, 289 passed, 27 expected failures
- independent adversarial review with fail-before/pass-after duplicate-ID coverage and current-main merge-tree proof
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Reorder editor tabs by dragging or using move-left/move-right commands.
  - Reorder sidebar panels by dragging or using directional commands.
  - Configure and persist sidebar panel order with `sidebar.panelOrder`.
  - Dragging tabs now supports edge auto-scrolling and visual drop indicators.

- **Bug Fixes**
  - Improved cancellation of drag interactions when views resize, hide, switch, or change.
  - Improved handling of wide-character tab labels and narrow layouts.

- **Documentation**
  - Added configuration details and examples for sidebar panel ordering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,6 +532,7 @@ To edit the raw files, use **Settings: Open settings.json** and **Settings: Open
 | `search.debounce` | int | `350` | Milliseconds to debounce global search input |
 | `explorer.showHidden` | bool | `true` | Show hidden files (dot-prefixed) in the file explorer |
 | `explorer.showGitIgnored` | bool | `true` | Show gitignored files in the file explorer |
+| `sidebar.panelOrder` | string[] | built-in order | Preferred sidebar panel IDs; updated when headers are dragged or moved with commands |
 | `terminal.shell` | string | `""` | Shell command for the integrated terminal (empty = system default) |
 | `terminal.scrollback` | int | `1000` | Number of scrollback lines to retain in the terminal |
 | `lsp.saveOnRename` | bool | `false` | Auto-save all files affected by a rename operation |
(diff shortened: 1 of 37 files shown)


arimxyer:
The merge-conflict suggestion above refers to the pre-rebase head. This branch is now rebased onto `fd06e36`, GitHub reports it as mergeable, and all CI checks are green.

eugenioenko:
@coderabbitai review

eugenioenko (commented):
_non blocking comment:_

There is this edge case: 
- User install History plugin that adds a new side panel.
- Re-orders it so that History is 1st
- Disables/Uninstalls the History plugin

The entry in `sidebar.panelOrder` persist.
In case of disabling, actually might be useful because after eneabling order persist.
In case of uninstalling, stale config.

It's one string, not a big deal. The only solution I can think of is after setting order, if any entry has not been used, to remove it from the order.  

Not a blocker, can be a follow up if at all.

eugenioenko (approved):
This is a really solid addition! Thank you for putting it together. Re-ordering side tabs is key for great UX and you've nailed it here. 

Nice!

PR #529, opened 2026-08-23
perf(render): avoid full-grid copy on every frame

### LINKED ISSUE #528: perf(render): Renderer.Render deep-copies the buffer every frame instead of swapping
## Problem

`internal/render/renderer.go:26-31` — the \"swap buffers\" step at the end of every `Render()` call fully reallocates and deep-copies the entire cell grid instead of swapping pointers:

```go
r.prev = make([][]term.Cell, len(r.curr))
for i := range r.curr {
    r.prev[i] = make([]term.Cell, len(r.curr[i]))
    copy(r.prev[i], r.curr[i])
}
```

Profiled with `make profiler` (cpu.prof/mem.prof): this is the single largest allocator in the app — 624MB flat / 973MB cumulative, 59% of all allocations captured in a session. It directly correlates with GC pressure in the CPU profile: even though the app was ~91% idle overall, 34% of actual on-CPU time went to `gcBgMarkWorker`/`gcDrain`.

Not a regression — present since the renderer was first introduced (commit 280eff7d, 2026-01-15).

## Fix

The caller (`internal/app/eventloop.go:155-159`, `redraw()`) already builds a brand-new `cells` grid every frame via `make()` and never reuses or mutates the previous one, so there's no aliasing hazard in keeping the reference instead of cloning:

```go
r.prev = r.curr
```

Replaces the alloc+copy loop with a single assignment.

eugenioenko:
## Summary
- `Renderer.Render` deep-copied the entire prev/curr cell grid into a fresh allocation on every single render call instead of swapping buffer references
- Profiling (`make profiler`, cpu.prof/mem.prof) showed this was the single largest allocator in the app — 624MB flat / 973MB cumulative, 59% of all allocations in a ~40s session — and a major source of GC pressure (34% of on-CPU time went to background GC marking despite the app being ~91% idle overall)
- Callers (`eventloop.go` `redraw()`) always pass a freshly allocated grid to `SetCurrent` each frame and never reuse/mutate the old one, so retaining the reference instead of copying is safe

Closes #528.

## Results
Before → after, comparable ~40s profiling sessions:
- Total heap allocations: 1639MB → 1245MB (**-24%**)
- `Renderer.Render`'s allocation cost dropped to zero (no longer appears in mem profile's top allocators at all)
- Remaining cost inside `Render` is 100% legitimate work (`screen.SetCell` + `screen.Show()`)

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` — all packages pass
- [x] `go test ./internal/render/... -v` — renderer unit tests pass
- [x] Re-profiled with `make profiler` after the fix and confirmed the allocation and CPU improvements above

## Summary by CodeRabbit

- **Performance**
  - Improved rendering efficiency by reusing the current grid instead of copying the entire grid after each render.
- **Bug Fixes**
  - Ensured rendered grid data remains correctly retained between rendering cycles.

### DIFF
--- a/internal/render/renderer.go
+++ b/internal/render/renderer.go
@@ -23,12 +23,10 @@ func (r *Renderer) Render(screen term.Screen) {
 		}
 	}
 	screen.Show()
-	// Swap buffers
-	r.prev = make([][]term.Cell, len(r.curr))
-	for i := range r.curr {
-		r.prev[i] = make([]term.Cell, len(r.curr[i]))
-		copy(r.prev[i], r.curr[i])
-	}
+	// No copy needed: callers always pass a freshly allocated grid to
+	// SetCurrent each frame, so retaining the reference is safe and avoids
+	// a full-grid alloc+copy on every render.
+	r.prev = r.curr
 }
 
 // Clear resets the renderer's buffers.
--- a/internal/render/renderer_test.go
+++ b/internal/render/renderer_test.go
@@ -3,6 +3,7 @@ package render
 import (
 	"github.com/eugenioenko/ttt/internal/term"
 	"testing"
+	"unsafe"
 )
 
 func makeCells(rows ...string) [][]term.Cell {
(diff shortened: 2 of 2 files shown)


eugenioenko:
@coderabbitai review

eugenioenko:
@coderabbitai review

Issue #524, opened 2026-08-23
fix(git): bound repository path identity retries during filesystem churn


arimxyer:
## Context

Follow-up hardening identified during review of #519.

`resolveRepositoryPathIdentityWith` has an outer retry loop with multiple `continue resolve` paths when a missing path or ancestor reappears during resolution. Under sustained filesystem churn, those restarts are not currently bounded. This is particularly risky for callers that resolve identity synchronously while preparing repository UI work.

## Desired invariant

A single identity-resolution request must finish after a bounded number of stability attempts. If the path keeps changing, return a descriptive transient error and fail closed; a later repository refresh may retry normally.

## Acceptance criteria

- Add a small, explicit retry cap covering every restart of the outer resolution loop.
- Preserve existing behavior for stable file and directory symlinks, deep missing paths, broken links, cycles, and permission failures.
- Add a deterministic injected `lstat`/`EvalSymlinks` test that would loop without the cap and proves bounded termination.
- Keep the change independent from repository refresh debouncing.

PR #530, opened 2026-08-23
perf(git): coalesce repository identity refreshes

### LINKED ISSUE #525: perf(git): coalesce active repository identity refreshes during watcher bursts
## Context

Follow-up hardening identified during review of #519.

`RepositoryState.InvalidateAll(RepositoryWorktree)` calls `refreshActiveIdentity()` before the existing worktree debounce. A burst of watcher invalidations therefore repeatedly cancels and launches identity reads even though the broader worktree refresh is already coalesced.

## Desired invariant

Watcher-triggered worktree bursts should produce one effective active-identity refresh per debounce window while preserving timely branch updates and stale-result safety.

## Acceptance criteria

- Coalesce active identity refreshes triggered by repository watcher invalidations.
- Preserve immediate refresh when the active file changes.
- Preserve periodic `HandlePoll` behavior that detects branch changes.
- Preserve cancellation, generation checks, last-good identity on transient failure, close behavior, and non-repository clearing.
- Add deterministic burst tests that count identity reads and prove the final repository state is current.
- Keep this independent from path-resolution retry bounding.

arimxyer:
## Summary

- debounce watcher-triggered active repository identity refreshes on a dedicated generation-checked trailing edge
- keep SetActiveFile, RefreshNow, and HandlePoll immediate while invalidating late watcher ticks
- explicitly rearm pending identity work across SetDirs and preserve cancellation, stale-result, last-good, close, and clearing behavior

## Invariant

A burst of debounced RepositoryWorktree invalidations starts zero identity reads before the latest main-loop debounce event and exactly one effective read after it. A later burst while that read is in flight waits for its own trailing boundary, cancels once, and allows only the final generation to apply. The identity timer callback only posts an event; repository state remains main-thread-owned.

The identity debounce is separate from the existing status debounce because an older in-flight status result may consume the status timer early. Keeping the timers separate avoids changing status follow-up and retry timing. The #524 path-resolution retry bound is unchanged.

## Verification

- Fail-before mutation: restoring eager watcher refresh made TestRepositoryInvalidationBurstCoalesces fail before the timer event with identity seq=3; current passes.
- go test ./internal/app -run focused-lifecycle-regex -count=100
- go test -race ./internal/app -run focused-lifecycle-regex -count=20
- go test -race ./internal/app -count=1
- go test ./... -skip exact-four-terminal-mouse-baselines -count=1
- go vet ./...
- make build
- git diff --check

The unfiltered all-package run passed every other package but hit the four existing 5-second terminal-mouse timing baselines: TestTerminalWidget_ForwardsSGRMouseClickAndRelease, TestTerminalWidget_ForwardsSGRWheel, TestTerminalWidget_CtrlClickStaysLocalEvenWithMouseReporting, and TestTerminalWidget_TrackingWithoutSGRStaysLocal. The exact-exclusion rerun passed.

Closes #525

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository identity updates during rapid file or directory changes by consolidating refreshes and avoiding unnecessary work.
  * Prevented outdated identity information from replacing newer results after overlapping refreshes.
  * Ensured identity data stays current when switching files, changing directories, or closing the repository.
  * Added safeguards so refreshes occur at the appropriate time without disrupting active operations.

### DIFF
--- a/internal/app/eventloop.go
+++ b/internal/app/eventloop.go
@@ -388,6 +388,8 @@ func RunEventLoop(
 				app.HandleDebugWriteRequest(v)
 			case *repositoryDebounceTick:
 				app.Repository.HandleDebounce(v)
+			case *repositoryIdentityDebounceTick:
+				app.Repository.HandleIdentityDebounce(v)
 			case *repositoryPollTick:
 				app.Repository.HandlePoll(v)
 			case *CommitLogResult:
--- a/internal/app/repository_state.go
+++ b/internal/app/repository_state.go
@@ -58,6 +58,7 @@ type RepositoryInvalidationRequest struct {
 }
 
 type repositoryDebounceTick struct{ Gen uint64 }
+type repositoryIdentityDebounceTick struct{ Gen uint64 }
 type repositoryPollTick struct{ Gen uint64 }
 
 type repositoryTimer interface {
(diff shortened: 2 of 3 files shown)


arimxyer:
@coderabbitai review

PR #538, opened 2026-08-23
plugin: API v3 gate + no-op Replace (prep for ttt-vim, 1.4.0)

eugenioenko:
Two changes for the ttt-vim compatibility effort:

1. **Bump `SupportedAPIVersion` to 3** so plugins can require the editor/API surface shipping with ttt 1.4.0 (the ttt-vim plugin now declares `"api": 3` — see eugenioenko/ttt-vim#7). Older builds already refuse unknown API versions via the existing `LoadManifest` check with a clear error.

2. **`PluginEditorAPI.Replace` no-ops when the replacement equals the range's current text.** Split `DeleteSelectionCommand` into `ComputeDeleted` (clamp + deleted-text computation, no mutation) and `Apply`; `Replace` skips the edit entirely on identical content. Calling `set_line` with unchanged text (e.g. the vim plugin's `=` reindent) no longer pushes a phantom undo group or triggers a redraw.

Tests: 5 new `ComputeDeleted` unit tests, 2 e2e tests (identical `set_line` doesn't consume an undo step; a real change does). `go test ./internal/core/undo/ ./internal/app/ ./tests/e2e/ -run 'TestPluginSetLine|TestUndoTo|TestDeleteSelection'` green.

## Summary by CodeRabbit

* **New Features**
  * Added support for version 3 plugin manifests, including updated editor settings and API capabilities.
* **Bug Fixes**
  * Avoided creating unnecessary editor changes or undo entries when replacement text matches the selected content.
  * Improved handling of deleted selections across lines, including boundary and cursor normalization.
* **Tests**
  * Added coverage for multi-line deletion behavior and plugin line-edit undo scenarios.

### DIFF
--- a/internal/app/plugin_api.go
+++ b/internal/app/plugin_api.go
@@ -206,6 +206,9 @@ func (e *PluginEditorAPI) Replace(startLine, startCol, endLine, endCol int, text
 		StartLine: startLine, StartCol: startCol,
 		EndLine: endLine, EndCol: endCol,
 	}
+	if delCmd.ComputeDeleted(ed.Buf) == text {
+		return
+	}
 	delCmd.Apply(ed.Buf)
 
 	if text == "" {
--- a/internal/core/undo/undo.go
+++ b/internal/core/undo/undo.go
@@ -441,9 +441,9 @@ type DeleteSelectionCommand struct {
 	Deleted             string
 }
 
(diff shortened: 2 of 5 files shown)


eugenioenko (commented):
This needs some more thought, it might be better to gate this as a specific compatibility setting instead of just doing it as core functionality

PR #557, opened 2026-08-24
fix(lsp): order signature-help requests after their triggering didChange

### LINKED ISSUE #556: ci(lsp test): flaky signature-help race,didChange can be sent after signatureHelp for the same keystroke
## Summary
`tests/functional/lsp-fake-protocol.test.js` ("sends signature help at the exact parenthesis position and renders its label") fails intermittently (~1-2% of runs locally, also seen on `main` post-merge CI: https://github.com/eugenioenko/ttt/actions/runs/32680420972/job/97296042563).

## Failure
```
AssertionError: expected [ Array(1) ] to deeply equal [ Array(1) ]
- Expected
+ Received
  [
    {
-     "text": "// fake protocol
console.log(",
+     "text": "// fake protocol
console.log",
    },
  ]
```
The `textDocument/signatureHelp` request is occasionally recorded before the `textDocument/didChange` for the triggering `(` keystroke.

## Root cause
`internal/app/app.go:517-519` calls `NotifyLSPChange(...)` then `CheckSignatureHelpTrigger()` synchronously, but each spawns its own independent goroutine to write to the LSP connection:

- `NotifyLSPChange` → `app_lsp.go:990` spawns a goroutine that calls `client.DidChange(...)`.
- `CheckSignatureHelpTrigger` → `RequestSignatureHelp` → `app_lsp.go:309` spawns a separate goroutine that calls `client.SignatureHelp(...)`.

Nothing orders these two goroutines against each other, so on rare scheduler timing the `signatureHelp` write can reach the wire before the `didChange` write for the same keystroke, violating the implicit ordering the LSP client (and this test) expects.

## Suggested fix
Combine the two into a single goroutine per keystroke so `didChange` is guaranteed to be sent before the `signatureHelp` request — e.g. one goroutine that calls `client.DidChange` first, then (if the trigger check passes) `client.SignatureHelp`, instead of two independently-scheduled goroutines.

## Reproduction
```
cd tests/functional
for i in $(seq 1 65); do npx vitest run lsp-fake-protocol.test.js --reporter=dot; done
```
Observed 1 failure in 65 local runs.

eugenioenko:
## Summary
- `NotifyLSPChange` and `RequestSignatureHelp` each dispatched to the LSP server from independent goroutines with no ordering guarantee between them, so on rare scheduler timing the `textDocument/signatureHelp` request could reach the server before the `textDocument/didChange` for the keystroke that triggered it.
- `NotifyLSPChange` now returns a `<-chan struct{}` that closes once its `didChange` notification has been sent; `CheckSignatureHelpTrigger`/`RequestSignatureHelp` wait on it before issuing the `signatureHelp` request, guaranteeing wire order without blocking the UI goroutine.

Fixes #556

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/app/...`
- [x] `cd tests/functional && npx vitest run lsp-fake-protocol.test.js` looped 80x — 0 failures (was ~1.5% flaky before the fix)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signature help accuracy after editing by ensuring the latest document changes are processed before suggestions appear.
  * Reduced timing-related issues that could cause signature information to be outdated or unavailable.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -514,9 +514,9 @@ func (a *App) Init(screen *term.TcellScreen, renderer *render.Renderer, lspManag
 			lang = a.EditorGroup.Editor.Highlighter.Language()
 		}
 		text := strings.Join(a.EditorGroup.Editor.Buf.Lines, "\n")
-		a.NotifyLSPChange(path, lang, text)
+		changeDone := a.NotifyLSPChange(path, lang, text)
 		a.ScheduleAutocomplete()
-		a.CheckSignatureHelpTrigger()
+		a.CheckSignatureHelpTrigger(changeDone)
 		a.ScheduleGitGutter()
 		if a.PluginManager != nil {
 			a.PluginManager.DispatchEvent("editor.change", path)
--- a/internal/app/app_lsp.go
+++ b/internal/app/app_lsp.go
@@ -260,12 +260,8 @@ func (a *App) charBeforeCursor() string {
 	return string(runes[col-1])
 }
 
-// CheckSignatureHelpTrigger inspects the character just typed and, if it's a
-// signature-help trigger, requests signature help. changeDone must be the
-// channel returned by the NotifyLSPChange call for this same keystroke, so
-// the request is sent only after that change has reached the server —
-// otherwise the server could see the signatureHelp request before the
-// didChange that produced the triggering character.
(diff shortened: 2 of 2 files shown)


eugenioenko:
@coderabbitai review

eugenioenko:
@coderabbitai review

PR #564, opened 2026-08-25
feat(changes): open unified diff view from root/section and PR root

eugenioenko:
## Summary
- Selecting a PR's root node in the Changes panel opens a new unified PR detail view: every changed file diffed in one scrollable `CommitDetailWidget`, matching how a single commit is shown. Since PR diffs are already fetched up front, this is built synchronously — no async read needed.
- Folder nodes keep their previous expand/collapse-only behavior.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/app/...`
- [ ] Manually verify: `ttt <pr-url>`, click the PR root node, opens the unified PR detail tab with all files diffed

## Summary by CodeRabbit

* **New Features**
  * Added navigation from the changes panel to pull request detail views.
  * Pull request details now open or focus in a dedicated tab.
  * Selecting a pull request displays its changed files and available diffs.
  * Added expandable sections in the changes tree for easier navigation.
  * Files can be opened directly from the changes tree to view their diffs.
  * Missing or empty diff information is surfaced within the pull request detail view.

### DIFF
--- a/internal/app/callbacks.go
+++ b/internal/app/callbacks.go
@@ -665,7 +665,6 @@ func registerWidgetCallbacks(app *App) {
 	}
 	app.Changes.OnOpenCommitDiff = app.OpenCommitDiff
 	app.Changes.OnOpenCommit = app.OpenCommitDetail
-	app.Changes.OnOpenCurrentChanges = app.OpenCurrentChanges
 	app.Changes.OnOpenPRDetail = app.OpenPRDetail
 	app.Changes.OnOpenPRDiff = func(group *ui.ChangesGroup, status git.FileStatus, extended bool) {
 		app.OpenPRDiff(group, status, extended)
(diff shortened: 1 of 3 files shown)


eugenioenko:
Open question for follow-up: should selecting the working-tree root/section node ("Staged (N)" / "Changes (N)" / a multi-repo group root) also open the unified Current Changes view, the same way the PR root does here?

I built and tested that variant, but held it back — the main tradeoff is that it changes what Enter/click does on that row (opens the view instead of toggling collapse; collapse/expand stays reachable via arrows, `h`/`l`, Space, and the chevron). Wanted to raise it explicitly rather than bundle it into this PR silently.

eugenioenko:
Addressed both CodeRabbit findings in 806ebbf:
- PR-root activation now only short-circuits when `OnOpenPRDetail` is actually set; otherwise it falls through to the normal expand/collapse toggle.
- A parsed diff with zero hunks is now treated as an empty-diff error, matching `OpenPRDiff`'s existing behavior.

PR #569, opened 2026-08-28
fix: allow opening files without automatic workspace creation

lecheel:
- Update resolveArgs to ignore openFiles when determining default folder
- Fix indentation in tcell_screen.go

Problem:
When launching the application with a file argument (e.g., ./ttt foo.go), the sidebar explorer was completely empty. However, running ./ttt with no arguments worked correctly and displayed the current working directory.

Root Cause:
In internal/app/widgets.go, the resolveArgs() function determines which folders to load into the workspace. It only added the current working directory (CWD) as a fallback if no files, directories, or PR URLs were provided. Because foo.go was added to the openFiles list, the CWD was skipped, resulting in an empty workspace and an empty explorer pane.

Fix:
Updated the condition in resolveArgs() to ignore openFiles when deciding whether to fallback to the CWD. Now, the app ensures the current directory is loaded into the workspace whenever no explicit directories or PR URLs are provided, regardless of whether specific files were opened via the CLI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Opening files explicitly no longer incorrectly adds the current directory as a workspace folder.
  * The current directory is used as a workspace only when no folders or pull request URLs are provided.

* **User Experience**
  * When files are opened without an explicitly selected folder or workspace, the sidebar is hidden by default to maximize editor space.

### DIFF
--- a/cmd/ttt/main.go
+++ b/cmd/ttt/main.go
@@ -359,8 +359,13 @@ Docs: https://tttedit.dev
 		w, h = flags.sizeW, flags.sizeH
 	}
 	editor.Root.SetSize(w, h)
-
 	editor.PendingFileTargets = fileTargets
+
+	// If the user only opened files and didn't explicitly open a folder/workspace,
+	// hide the sidebar by default to maximize the editor space.
+	if !editor.ExplicitFolders && len(fileTargets) > 0 {
+		editor.HideSidebar()
+	}
 
 	if flags.pluginFile != "" {
 		app.LoadPluginFromFile(editor, flags.pluginFile)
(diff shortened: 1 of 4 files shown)


eugenioenko:
Hi @lecheel, thank you bringing this up!

> It only added the current working directory (CWD) as a fallback if no files, directories, or PR URLs were provided.

This was  kinda intentional. When opening a single file the workspace folder is intentionally not included because when a filename passed it's assumed the intent is to open that file and nothing else.
If you want to open a workspace and a file, it's possible to do `ttt . README.md` which would open the current folder and the file.

I think at the implementation time, it was a workaround so that the sidebar wouldnt open when opening one file.

That's why a few of functional tests are failing probably as well. Loading the directory makes sense,  the sidebar panel should stay closed when opening a single file and not a directory though. Because intent was to open that single file.

Can you please update PR to keep sidebar closed? That should fix the failing  tests

PR #591, opened 2026-09-07
fix: ignore whitespace in Quick Open files

vikas-0:
## Summary

Normalize Quick Open queries by removing whitespace before fuzzy matching.

This allows queries such as `user profile` to match filenames like
`user_profile.go` and `user-profile.go`. Previously, spaces in the query were
treated as literal characters, making matching asymmetric: `userprofile`
matched `user_profile`, but `user profile` did not.

## Testing

- Added regression coverage for whitespace-separated file queries
- `go test ./internal/ui`

## Summary by CodeRabbit

* **Bug Fixes**
  * File search now ignores spaces and other whitespace in queries, improving fuzzy matching for filenames such as `user_profile.go` and `user-profile.go`.

* **Tests**
  * Added coverage to verify whitespace-separated searches return the expected files.


eugenioenko (approved):
@vikas-0  thank you for creating a PR to fix this. Great catch! 

Thank you for helping making ttt better, really appreciate it :)

PR #599, opened 2026-09-11
highlight: external chroma lexers, and multi-line regions beyond block comments

austinm911:
Draft: two related changes to `internal/highlight`, sitting behind one motivating case. Happy to split them into separate PRs if you'd rather review them apart.

I wanted TSRX (`.tsrx`, https://github.com/tsrx-org/tsrx) highlighted in ttt. Chroma does not ship a lexer for it, the Lua plugin API has no highlighting hook, and a language cannot be added without a Go change. These two commits fix the general problem rather than adding one more language.

## 1. External chroma lexers

`~/.config/ttt/lexers/*.xml` is scanned once at the first `highlight.New`. Each file is a serialised chroma lexer, registered into chroma's global registry, so every surface that already calls `highlight.New` picks up the language by filename: editor tabs, diff views, commit detail, readonly tabs, markdown fences. A lexer next to the binary can be overridden by one in `~/.config/ttt`, matching how themes resolve.

```sh
mkdir -p ~/.config/ttt/lexers
cp tsrx.xml ~/.config/ttt/lexers/
bin/ttt app.tsrx     # status bar reads TSRX
```

Two robustness points, both learned the hard way while writing a lexer:

- Chroma **panics**, not errors, on a missing `using` delegate or a bad state name. Files are probed before registration, `lexLine` recovers, and a rejected file is reported in the Output panel at startup instead of failing silently.
- The registry must be attached with `SetRegistry` before probing, or `<using lexer="..."/>` panics.

`cmd/lexcheck` is the tool I used to develop the lexer. It drives the real `highlight.HighlightLineAt`, one line at a time, with region state carried, so it sees exactly what the editor sees:

```sh
bin/lexcheck dump app.tsrx                  # every span, line by line
bin/lexcheck check app.tsrx                 # asserts app.tsrx.assert
# assertion format: <line>|<substring>[#occurrence]|<style>
```

If you would rather not carry another `cmd/`, say so and I will move it under `internal/highlight` as a test helper.

## 2. Multi-line regions, not just block comments

The existing cross-line state was one bool and one delimiter pair, so a block comment survived a newline and nothing else did. A template literal, a Go raw string or a Python docstring lost its color on every line after the first.

That is now a list of regions discovered by probing the lexer, with the same memoized state table, the same invalidation, and the same "no full-buffer re-lex" rule. `blockOpen`/`blockClose` become `regions []region`; `states []bool` becomes `states []int8`.

```
const sql = `        before          after
  SELECT 1           plain           string
`;                   plain           string
```

Detection needs both views of the lexer, which is the one subtle part. Chroma coalesces adjacent tokens of the same type, so `""" a\nb """` reads as a single string token in Go and Rust, neither of which has a triple-quoted literal; they see `""`, `" a\nb "`, `""`. Requiring the **raw** lexer to emit the delimiter as a token of its own separates a real region, where the opener pushes a state, from three literals in a row. Markdown, JSON, YAML and TOML come out with no string region, Go gains its raw string, Python its docstrings, JS/TS their template literals.

### Performance

`AGENTS.md` calls full-buffer re-lexing a trap, so the per-line cost was measured rather than assumed. A one-pass `ContainsAny` over the delimiters' leading bytes rejects nearly every line before any lexing, and `closesAt` no longer materialises the line as `[]rune`.

| benchmark | before | after |
| --- | --- | --- |
| `HighlightLine/cold` | 86.1 µs | 78.2 µs |
| `HighlightLine/warm` | 11.4 ns | 12.4 ns |
| `Invalidation/200000/edit=start` | 1.86 ms | 1.87 ms |
| `Invalidation/200000/edit=end` | 281 µs | 272 µs |

## Tests

`internal/highlight` is green, including new cases for template literals across lines, Go raw strings, Python docstrings, a closed template not leaking state into the next line, escaped closers, earliest-region-wins, and two regression tests for the coalescing false positive (`""" ... """` in Go/Rust/JS, backticks in Markdown).

The pre-existing `internal/app` failures under `go test ./...` on macOS (`/private/var` vs `/var`, and the fake-git process tests) reproduce unchanged on a clean `origin/main` checkout here.

## Why not upstream the lexer to chroma instead

For TSRX specifically I intend to do that too. It does not replace this: a chroma release plus a ttt chroma bump is the cycle for every new or private language, and this makes the local case a file copy.


austinm911:
Closing in favour of two smaller PRs, one per change, each green on `main` on its own.

- #600 `fix(highlight): keep multiline string color across lines` — the multi-line regions half, extending #441. Retitled to what a user sees, and the description now states what it does not cover.
- #601 `feat(highlight): load external chroma lexers from config lexers/ dirs` — the loader half, with three guards this PR was missing.

What changed beyond the split, since reviewing my own branch turned up more than the split did:

- the loader commit here never worked on its own. The only trigger for the one-shot scan was in the *other* commit, so at `a4b2044` nothing registered and two of the three tests it adds fail. #601 owns its trigger and its `recover`.
- an external lexer could silently replace a builtin one. chroma's `Register` replaces by `<name>`, so a file named `Go` killed highlighting for every `.go` file with no message. #601 rejects name collisions, duplicate names within a scan, and oversized files.
- the probe only reaches the `root` state, so it passed lexers that panic on real input. #601 adds a static pass over every rule's `Push` target and stops claiming the probe is a gate.
- regions had four over-colouring regressions against `main`. Go raw strings were assumed to honour backslash escapes, so `C:\` before the closing backtick left the region open for the rest of the buffer; collecting every comment candidate gave TypeScript an `<!-- -->` region, so `a <!--b` commented out the rest of the file. Both are fixed in #600 by asking the lexer instead of assuming, with a regression test each.
- the performance table here was noise in both directions. Re-measured interleaved against `main`: no measurable change, allocations identical.

`cmd/lexcheck` is in neither PR. The release workflow builds `./cmd/ttt` only, so it never reached a user, and the assertion corpus it drives is for a specific language and belongs outside this repo.

PR #600, opened 2026-09-11
fix(highlight): keep multiline string color across lines

austinm911:
Extends #441 to the rest of the constructs that span lines.

## The bug

`#441` gave the highlighter one bool and one delimiter pair of cross-line state, which is exactly what a block comment needs. Nothing else carries. On `main`:

```
                 line          spans
ts-template   1| const q = `   keyword "const"  string "`"
              2| SELECT 1      number "1"          <- lost the string
              3| `;            string "`;"
go-rawstring  2| SELECT 1      number "1"          <- lost the string
py-docstring  3| doc           (no span at all)    <- lost the string
block-comment 2| comment       comment "comment"   <- the case #441 fixed
```

A template literal, a Go raw string and a Python docstring each keep their colour on the opening line and lose it on every line after.

## The fix

The pair becomes a list of regions, discovered the way #441 discovered the comment pair: by probing the lexer, never by matching delimiter text. Same memoized state table, same invalidation, same rule against re-lexing the buffer. `blockOpen`/`blockClose` become `regions []region`; `states []bool` becomes `states []int8`.

Two things the probe decides rather than the candidate table:

**A language has one block comment.** The first comment candidate that matches wins. Collecting every match gives TypeScript an `<!-- -->` region on top of `/* */`, and `if (a <!--b) { c() }`, which is valid TypeScript, then comments out the rest of the file.

**Whether a backslash escapes the closing delimiter is a per-language fact.** It is true for a JavaScript template literal and false for a Go raw string. Hard-coding it either way breaks one of them, and the Go direction is unbounded: `ttt -d C:\` before the closing backtick leaves the region open for the whole buffer. The probe appends a sentinel after the closer and asks whether one token covers it.

## Where detection stays conservative

Chroma coalesces adjacent tokens, so `""" a\nb """` reads as one string in Go and Rust, neither of which has a triple-quoted literal; they see `""`, `" a\nb "`, `""`. Requiring the raw lexer to emit the delimiter as its own token separates a real region from three literals in a row.

That test is conservative, not exact. **Swift, Scala and Java do have multi-line `"""` strings and are rejected by it** (Kotlin passes and gains the region), so they keep `main`'s behaviour rather than gaining a wrong one. Swift's and Rust's token streams for the probe are byte-identical, so no refinement separates them; admitting Swift means admitting Rust, where the same input is three ordinary literals. Markdown, JSON, YAML and TOML gain no string region. Python gains both `"""` and `'''`.

## Known limits, both pre-existing in shape

`HighlightLine` still ignores regions, so diff views, commit detail and the markdown renderer are unchanged by this PR. Only the editor pane calls `HighlightLineAt`.

One input still over-colours: a Python docstring line ending in a backslash, `x = """a\` then `b"""`, is read as opening nothing and then opening a region on the closing line. Deciding it correctly needs the lexer's state at the end of line one, which chroma v2 does not expose; `TokeniseOptions.State` is a single initial state name and the final stack is not returned.

## Performance

No measurable change. Interleaved runs against `main` to cancel machine drift, `-benchtime 2000x`:

| | main | this PR |
| --- | --- | --- |
| `HighlightLine/cold` | 78.1–80.3 µs | 74.7–77.1 µs |
| `HighlightLine/warm` | 11.8–14.4 ns | 11.7–14.5 ns |

Allocations are identical on both branches: 38624 B/op and 599 allocs/op cold, zero warm. `BenchmarkHighlightLineAtAfterInvalidation` at 200k lines is unchanged.

## Tests

`internal/highlight` is green. New cases: a template literal across lines, a Go raw string, a Python docstring, a closed template not leaking state into the next line, an escaped closer, earliest-region-wins, the coalescing false positive for Go/Rust/JS and for Markdown backticks, plus one regression test each for the two probe decisions above.

## Summary by CodeRabbit

* **New Features**
  * Improved syntax highlighting for multi-line comments, strings, templates, raw strings, and docstrings.
  * Preserved coloring across multi-line regions and closing delimiters.
  * Added support for escaped delimiters and overlapping region handling.

* **Bug Fixes**
  * Improved highlighting accuracy across supported languages.
  * Prevented incorrect region tracking when lexer detection fails.

* **Documentation**
  * Clarified how multi-line syntax regions are detected and tracked.


eugenioenko:
Reviewed for correctness and performance.

**Performance:** no regression. Benchmarked `HighlightLine` (cold/warm) and `HighlightLineAtAfterInvalidation` against `main` on the same machine — numbers match this PR's own claims, and the extra lexer probing in `detectRegions` only runs once in `New()` at file-open, never in the per-line hot path.

**@coderabbitai's `'''` finding (line 354) is correct** and should be fixed before merge:
```go
{open: ", close: ", style: term.StyleSyntaxString, ownsDelim: true},
```
`close` is left as the zero-value `""` and `open` is a garbage string no lexer will ever emit, so Python `'''`-style docstrings (the common style, not just `"""`) get no multi-line region at all — the exact bug this PR fixes, still present for this delimiter. No test exercises `'''` (`TestPythonDocstringSpansLines` only covers `"""`), which is how it shipped unnoticed. Suggested fix: `{open: "'''", close: "'''", ...}` plus a `'''` case alongside the existing docstring test.

**@coderabbitai's `opens` memo nitpick (lines 62-64) is accurate but pre-existing** — `main` already never clears this map in `ClearCache`, this PR just changes its value type. Not a regression from this PR; no action needed here.

**New gap found in review, filed separately:** PHP docblocks (`/** ... */`) don't get multi-line comment tracking either — pre-existing on `main`, not a regression, but worth tracking since this PR generalizes the mechanism. #602

austinm911:
Fixed in 110b135. The finding is correct and it is mine, not a subtlety: I wrote that table through a Python heredoc and its quoting ate the delimiters, leaving `{open: ", close: ", ...}`, which Go reads as an open of `", close: "` and an empty closer. It compiles, which is why it shipped.

Two tests, both failing before the fix and passing after:

- `TestPythonSingleQuoteDocstringSpansLines`, the behavioural one you asked for, alongside the `"""` case.
- `TestRegionCandidatesHaveBothDelimiters`, because an empty closer is worse than a missing region: `closesAt` returns -1 for it, so had the probe ever matched, the region would have run to the end of the buffer. No language matched it in practice, but the table should not be able to express it.

Detected regions after the fix, measured across 28 extensions:

```
py     ["""..."""] ['''...''']      go   [/*...*/] [`...`]
js/ts  [/*...*/] [`...` esc]        rb   [`...`]
jl     ["""..."""] [`...`]          kt   [/*...*/] ["""..."""]
lua    [--[[...]]]                  hs   [{-...-}]
ml     [(*...*)]                    ps1  [<#...#>]
html   [<!--...-->]                 md/json/yaml/toml/sh/ex  none
```

That run also corrected the PR description: I had listed Kotlin among the languages whose `"""` the coalescing test rejects. It does not, Kotlin passes and gains the region. Swift, Scala and Java are the rejected ones. Body updated.

On the `Jungle` hang, your and CodeRabbit's read is right, and I confirmed it is not this PR:

```
$ timeout 40 go test ./internal/highlight -run TestZZJungleProbes -v
    probing "/*" ...        <- never returns
```

It is the first probe, `/* a\nb */`, which is byte-identical to `blockCommentCandidates[0]` on `main`. `origin/main` hangs the same way on `New("a.jungle")`, so the exposure is unchanged by this PR: same first probe, same lexer, same regexp2 backtrack. Worth its own issue if a user ever opens a `.jungle` file, since it hangs at file open rather than degrading.

On #602, your diagnosis holds against the code as written. `trailingRegionStart` gates on `tok.Type == r.tokenType`, and chroma's PHP lexer disagrees with itself about the two forms:

```
/* a\nb */   -> CommentMultiline
/** a\nb */  -> LiteralStringDoc
```

so the docblock opener is never recognised as the region's opener. I agree it should not be patched by widening `isCommentToken`, for the reason you give.

`opens` is left alone.

eugenioenko:
@coderabbitai full review

eugenioenko:
Nice fix for the multi-line tracking gap — but it introduces a regression worth fixing before merge.

**Bug:** `computeOpensAt`/`trailingRegionStart` (`internal/highlight/highlighter.go`, ~L367-410) misfires when a line's *own* closing delimiter is the last thing on the line. The probe appends a synthetic closer to test for an open region; when the real closer is already at end-of-line, the synthetic one coalesces into the same token, so the token still starts before `lineLen` and passes the `start < lineLen` check — the line gets wrongly flagged as opening an unclosed region.

Repro (verified against this branch):

```go
lines := []string{"const a = `x`", "const b = 2;"}  // JS, no trailing `;` after the backtick
lines := []string{`x = """abc"""`, `y = 2`}          // Python
```

In both cases the next line's first token comes back styled as a string instead of a keyword — the region state leaks past a line that already closed cleanly.

This is broader than the documented backslash-EOL limitation: it hits any closed one-line template/raw-string/docstring that isn't followed by more tokens on the same line — e.g. `` const a = `sql` `` with no semicolon, `q := \`select 1\`` as a Go statement, `x = """text"""` as its own statement. `TestClosedTemplateDoesNotOpenRegion` doesn't catch it because it has a trailing `;` after the backtick, which breaks the coalescing that triggers this.

Before this PR, strings had no cross-line tracking at all, so this specific case couldn't leak — it worked by omission. Might be worth a `start >= lineLen` check that also accounts for the closer's own length, or asserting the last matching token doesn't itself contain the appended closer's full text.

austinm911:
Fixed in 0161a53. The finding is right and the cause is exactly as you read it: the appended closer coalesces with the closer the line already had, so the merged token still starts before `lineLen` and the line is flagged as opening.

Two of the three repros reproduce; the Go one does not:

```
const a = `x`      -> leaked (opensAt col 10)
x = """abc"""      -> leaked (opensAt col 4)
x = '''abc'''      -> leaked
q := `select 1`    -> no leak
x = --[[ a ]]      -> no leak
```

Go and Lua were already fine: their lexers do not coalesce the appended closer into the same token, so `trailingRegionStart` returned -1 there. The leak is specific to lexers that do.

Your suggested lever, accounting for the closer's own length, does not separate the cases: `` `x` `` closed and `` `x` `` + one appended char differ by exactly that length, so any length arithmetic gets one of them wrong. The root problem is that `computeOpensAt` used the lexer to answer two different questions, and closure was one of them. `closesAt` already answers closure for `nextState` and `computeSpans`, so it now answers it here too:

```go
start := h.trailingRegionStart(line+r.close, r)
after := start + len([]rune(r.open))
if start < 0 || after > len(runes) {
    continue
}
if closesAt(string(runes[after:]), r) >= 0 {
    continue
}
```

The lexer is asked only where a region starts, which is what it is needed for: an opener inside a string or after a line comment is not one. Closure has a single oracle across opens, state and spans, so the two can no longer disagree whatever a lexer coalesces.

The `after > len(runes)` guard is the second half, and it closes a hole the old `start >= lineLen` check left. A line ending in `""` plus the appended `"""` synthesizes an opener the line never had. It happens to be rejected today only because chroma types a short `""` as `LiteralStringDouble` rather than the region's `LiteralStringDoc`, which is luck, not a rule. Requiring the opener to fit inside the original line makes it structural.

Tests, the first two failing before the fix and passing after:

- `TestRegionClosedAtEndOfLineDoesNotLeak` — js, go, lua, py `"""` and py `'''`, all with the closer at end of line. The existing `TestClosedTemplateDoesNotOpenRegion` keeps its trailing `;`, which is what broke the coalescing and hid this.
- `TestAppendedCloserCannotCompleteAnOpener` — `x = ""`, `` const a = `` ``, `code(); /`.
- `TestRegionReopenedAfterClosingOnSameLine` — passes both before and after, and exists to catch the wrong fix: rejecting any line that contains a closer. `` `a` `b` `c ``, `/*a*/ /*b*/ /*c` and `''' ... ''' + """` must still carry.

Also verified against over-rejection, 37 cases: cross-region shadowing (`` /* `x */ ``, `` `a /* b ``, `` // `a ``, `` "`" ``) and escapes both ways (`` `a\` `` opens in JS, `` `a\\` `` does not, Go `` `a\` `` does not). One divergence remains and is inherent to a textual closer: `` `x${`y`}z `` reads as closed. `nextState` would end the region at the same backtick, so `opensAt` and the state table agree, which is the property the fix is for.

Benchmarks unchanged. `BenchmarkHighlightLineAtAfterInvalidation` is the path that calls `opensAt` per line, and it is within 1% of pre-fix at 50k and 200k lines, edits at start and end. The added work is one `closesAt` scan per candidate on a cache miss, against a lexer pass that was already there.

On @coderabbitai's other finding, the assignment-style Python opener at L327-329: not reproducible. The opener is found in expression position.

```
"x = \"\"\"a"   -> {col:4 region:0}
"\"\"\"a"       -> {col:0 region:0}
"    \"\"\"a"   -> {col:4 region:0}
```

Chroma's Python lexer emits `LiteralStringDoc` for `"""` in both positions, which is the type the probe recorded, so `trailingRegionStart` matches. No change made.

Its test-matrix ask was fair and is now `TestTripleQuoteRegionLanguageMatrix`: py and kt gain the `"""` region, swift, scala, java, go and rs do not. That pins the matrix in the PR description rather than leaving it as prose.

eugenioenko (approved):
@austinm911 thank you for working on this fix for multiline string detection. With ttt and chroma constraints, this is genuinely complex to nail down. Really appreciate your effort here.

Thank you for making ttt better for the community!

PR #610, opened 2026-09-15
feat: welcome screen over pristine scratch tabs

sjdonado:
Centered welcome text instead of an empty editor buffer on fresh starts and close-all, Vim-style.

## Summary

image

- When the only tab is a pristine (empty, clean) scratch buffer, the editor area shows the welcome text above instead of the empty buffer. The tab bar still shows `untitled`; save flow unchanged.
- Typing takes over instantly (first keystroke dirties the buffer, overlay vanishes); cursor hidden while showing.
- Opening a real file consumes the pristine scratch tab instead of piling up next to it; a dirtied scratch buffer is never replaced, and pinned slots are never consumed. Explicit New File still appends.
- Closing all tabs lands back on welcome. Read-only openers consume pristine tabs the same way; diff and plugin tabs are untouched.

## Test plan

- [ ] Fresh start with no files: centered welcome with version and hints, no cursor
- [ ] Type anything: editing starts immediately, welcome gone
- [ ] Open a file from welcome: single file tab, no leftover untitled
- [ ] Type in scratch, then open a file: scratch tab kept
- [ ] Close all tabs: welcome returns
- [ ] `go test ./internal/ui/ ./tests/e2e/`

## Notes

- Rendered as an overlay: the editor stays laid out underneath, so the first keystroke can never observe a zero-width viewport (caught by an existing listen test during review).
- Clicks on the welcome report no position, so context menus skip buffer-dependent plugin items; signature/autocomplete/hover layers stay hidden while it shows.
- Full e2e suite green; `internal/app` (3) and `internal/ui` (4) failures reproduce identically on clean HEAD (macOS `/var` symlink + PTY env issues); functional `listen-shutdown` also fails with a pristine-base binary. All unrelated.

## Checks

`go build`, `go vet`, `gofmt` clean; predicate unit test, 5 welcome e2e tests (shows, typing, new-file, open-consumes, close-all-returns, dirty-kept), pin/tab-reorder updates for the new tab behavior, and the full Go + functional suites are green; remote CI is green on this PR.

## Assumptions

- Welcome content (title + 4 hints) is a starting point; wording is cheap to change if you want different hints.
- Pinned pristine tabs are never consumed (avoids corrupting the pinned region); an edge case that trades one stranded scratch tab for pin integrity.
- Diff and plugin tabs never consume or trigger welcome; only file openers do.
- The overlay repaints the editor each frame while showing (idle pristine state only); accepted over risking stale editor state.

## Rejected review findings

- Removing the untitled tab model instead: an order of magnitude more churn across ~10 test files for no UX gain; the overlay is indistinguishable to the user.
- Undo-history in the pristine predicate: undo never clears the dirty flag (only save/load do), so typed-then-undone buffers already read non-pristine.
- Whole-line copy/cut/select-all guards: identical behavior on any empty buffer, not welcome-specific.
- `tabW <= 0` guards: `resolveTabSize` guarantees ≥ 1 at the single choke point, same as existing unguarded divisions.
- Consuming all pristine tabs on open (multi-scratch piles): only the active one is consumed; deliberate Ctrl+N spam leaving one empty scratch is a non-issue for now.
- Partial theme overrides and per-field fills: whole-struct layering is the convention (no theme changes in this diff anyway).

## Summary by CodeRabbit

- **New Features**
  - Added a welcome screen for fresh editor sessions, including version information, keyboard hints, and common actions.
  - Welcome content disappears when typing or creating a new file and returns after all tabs are closed.
  - Welcome overlays suppress editor popups while displayed.

- **Bug Fixes**
  - Opening files now reuses a fresh untitled tab without affecting pinned tabs.
  - Dirty scratch tabs remain available when opening files.
  - Read-only files and buffers follow the same tab reuse behavior.


eugenioenko (commented):
Hi @sjdonado, thank you for contributing!

> centered welcome text instead of an empty editor buffer on fresh starts and close-all, Vim-style.

Vim certainly needs the feature because it's really not intuitive for new users.
VSCode when no tabs are open has the background centered text with similar info. I'm not sure at this point how helpful it is.

This PR introduces a similar screen to VSCode which helps with some shortcut discoverability, that's great. But a few things are not obvious and confusing that need to be fixed before this feature could be merged:

- Clicking the untitled tab buffer does not focus the cursor on the editor, you need to type for the welcome message to be dismissed. Meaning, there is no indication that you can start typing.
- Closing the tabs and getting a new untitled tab, you get the message again. Which is similar to VSCode when all tabs closed, and in ttt you can't close tabs. But if it's a welcome screen, shouldn't it be shown once?

On more of a discussion note:
This does help with discoverability for some important shortcuts. A question I can't answer is "How much does it really help?" Meaning, opening the View/File menu gets you the same shortcuts.

So the question is: who is this feature for?

Addressing the two UX frictions would make the introduction land smoothly. It would be good to merge. But I'd like to talk through *why* we're saying yes to this before merging.

sjdonado:
> Hi @sjdonado, thank you for contributing!
> 
> > centered welcome text instead of an empty editor buffer on fresh starts and close-all, Vim-style.
> 
> Vim certainly needs the feature because it's really not intuitive for new users. VSCode when no tabs are open has the background centered text with similar info. I'm not sure at this point how helpful it is.
> 
> This PR introduces a similar screen to VSCode which helps with some shortcut discoverability, that's great. But a few things are not obvious and confusing that need to be fixed before this feature could be merged:
> 
> * Clicking the untitled tab buffer does not focus the cursor on the editor, you need to type for the welcome message to be dismissed. Meaning, there is no indication that you can start typing.
> * Closing the tabs and getting a new untitled tab, you get the message again. Which is similar to VSCode when all tabs closed, and in ttt you can't close tabs. But if it's a welcome screen, shouldn't it be shown once?
> 
> On more of a discussion note: This does help with discoverability for some important shortcuts. A question I can't answer is "How much does it really help?" Meaning, opening the View/File menu gets you the same shortcuts.
> 
> So the question is: who is this feature for?
> 
> Addressing the two UX frictions would make the introduction land smoothly. It would be good to merge. But I'd like to talk through _why_ we're saying yes to this before merging.

Good point, thank you for reviewing. It just felt unfamiliar to me to always have the Untitled tab there. This is more like a workaround, but I think now that a real solution would be not to have an empty buffer automatically opened. This is a bigger question, and I don’t have a good proposal for it. I will close this PR so it can be discussed in an issue thread instead.

PR #615, opened 2026-09-17
fix(terminal): detect shell exit by reaping the child, not by reading the pty

SimonOcampo1:
Typing `exit` in the integrated terminal left a dead terminal on screen: the tab stayed, and so did the bottom panel. Only `Ctrl+T` or `Terminal: Close All` got rid of it.

## Cause

`OnExit` never fired.

`readLoop` treats a failed read on the pty master as the signal that the child is gone. That never happens here: go-pty assigns the slave to the child's stdio and keeps its own copy of that fd open in this process, so the master never reports EOF when the child dies, and the read blocks forever.

```go
cmd.Stdin  = pty.slave
cmd.Stdout = pty.slave
cmd.Stderr = pty.slave
```

With the slave still open in the parent, no amount of waiting on the read will do.

## Fix

Wait on the process instead. `waitLoop` reaps the child, marks the terminal exited, closes the pty — which is what finally releases `readLoop` — and then fires `OnExit`. `Close()` no longer runs a competing `Wait()`; it kills and waits for the reaper through a channel, so the two paths cannot race.

`CloseTerminal` now also collapses the bottom panel when the last terminal goes away, but only when the terminal is the tab actually being shown — the panel also hosts Problems, References and Output, and closing it out from under someone reading diagnostics would be worse than the bug.

## Tests

Two regression tests, both directions: `OnExit` fires when the shell exits, and does not fire on an explicit `Close()`.

`go build ./...`, `go vet ./internal/...` and `go test ./...` pass, including `-race` on `internal/terminal`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal lifecycle handling so exit notifications occur reliably when a shell process ends.
  - Prevented terminal cleanup from triggering duplicate exit notifications when a terminal is closed manually.
  - Closing the last terminal now collapses the bottom panel only when the terminal view is active.
  - The bottom panel remains open when displaying Problems, References, or Output, while focus returns to the editor.


eugenioenko (approved):
@SimonOcampo1 thank you for catching this! Solid bugfix :)

PR #616, opened 2026-09-17
fix(terminal): read the cursor from the same snapshot as the cells

SimonOcampo1:
A TUI running inside the integrated terminal paints its prompt, then draws the cursor somewhere else and leaves it there. Resizing the window — anything that forces another frame — puts it back. Reproduced with Claude Code running inside ttt.

## Cause

`Render` takes a `Snapshot` under the emulator lock to read the cells. `CursorPosition` then takes the lock **again** and reads `Term.CursorPos()`. The two run against different emulator states, so a frame can pair this frame's content with a later cursor.

Whether it looks wrong depends on how fast the child is repainting, and once the child settles nothing forces a correcting frame — which is why the cursor stays misplaced until an unrelated event arrives.

## Fix

Capture cursor position and visibility inside the render snapshot, and return those from `CursorPosition`.

This also makes the widget honour `CursorVisible()`, which it ignored: a program that hides the cursor while repainting no longer leaves one parked at a stale position.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cursor position reporting now stays synchronized with the terminal content rendered in each frame.
  * Invisible cursors are no longer reported as visible at position 0,0.


eugenioenko (approved):
@SimonOcampo1  Nice catch! Thank you for helping making ttt better for the community :)

PR #617, opened 2026-09-17
feat(panel): let the bottom panel dock to the right edge

SimonOcampo1:
The panel that holds Terminal, Problems, References and Output is wired to the bottom. On a wide screen a terminal 13 rows tall wastes the width the editor does not need, and there is no way to move it — the palette only offers `Increase/Decrease Panel Height`.

Adds `View: Move Panel Right` and `View: Move Panel Bottom`, persisted as `panel.position` so the choice survives a restart.

## Shape of the change

`ContentSplitWidget` gains a `Position` whose zero value is `SplitBottom`, so every existing instance — the commit-history split among them — keeps the horizontal layout without opting in. `Render` and `HandleEvent` branch on it, and the divider drags along whichever axis is active.

`RightW` is a separate field from `BottomH` on purpose: 15 rows is a reasonable height and a useless width, so sharing one field hands you a 15-column panel the first time you move it. `SpawnTerminal` derives cols and rows from whichever axis is active — 111x13 docked at the bottom, 57x25 docked right, on the same window.

`View: Increase/Decrease Panel Height` are now `... Panel Size`, since they grow the panel along whichever axis it currently uses. Command IDs and key bindings are unchanged.

## A settings bug this uncovered

`panel.position` saved once and never again: the struct held the new value at marshal time and the file came back with the old one.

`Settings.MarshalJSON` captures top-level keys absent from `knownSettingsKeys` into `Extra` — meant for plugin-namespaced settings — and reinjects `Extra` over the struct encoding on save. A new core key that is not registered there round-trips through `Extra` and overwrites itself with the stale value on every save after the first.

Registering `"panel"` fixes it. Since this will bite the next person who adds a settings key, there is also a reflection test that fails for any field of `Settings` whose JSON name is missing from that map. I checked that it fails without the one-word fix.

## Summary by CodeRabbit

- **New Features**
  - Dock the bottom panel along the bottom or right edge of the window.
  - Move the panel between positions using the panel controls.
  - Resize the panel appropriately for its current position.
  - Panel placement and size are retained between sessions.
  - New terminals automatically use the available space based on the panel layout.

- **Bug Fixes**
  - Improved panel resizing and layout behavior when switching dock positions.


eugenioenko:
@coderabbitai reivew

eugenioenko (commented):
Hi @SimonOcampo1 thank you for working on this!  With wide monitors it makes a lot of sense to have the terminal on the right :) 


This is mostly ready to go, one small change I'll ask is to rename the Command Pannel action to ` View: Dock Panel Right / View: Dock Panel Bottom`

There is already a command to move sidebar panel to the right and its confusing a little bit,  docking fits better here as well.


And if you get a chance, and no need to do it (I can follow up with that later) would you be able to adjust the corner icon? 

When the terminal is at the right, the top-right edge corner `╮` is misplaced:

```
  File   Edit   Selection   View   Options   Help
╭────────────────────╮│ Terminal  Diagnostics  References  Output  +  ⋮
│ Explore       »  ⋮ ││─────────────────────────────────────────────────
│────────────────────││[>1]│ …/enko/Documents/ttt [feat/movab          █╮
│▼ ttt              █││    │ …/enko/Documents/ttt [feat/movab          █│
│  ▶  .claude      █││    │ …/enko/Documents/ttt [feat/movab          █│
│  ▶  .git         █││    │ …/enko/Documents/ttt [feat/movab          █│
│  ▶  .github      █││    │ …/enko/Documents/ttt [feat/movab          █│
│  ▶  audit        █││    │ …/enko/Documents/ttt [feat/movab          █│
╰────────────────────┴──────────────────────────────────────────────────╯
                                 Ln 1, Col 1   Spaces: 4   UTF-8   LF
```

SimonOcampo1:
Thanks! Done:

- Renamed to `View: Dock Panel Right` / `View: Dock Panel Bottom` (command IDs `panel.dockRight` / `panel.dockBottom`, and the ⋮ menu reads "Dock Right" / "Dock Bottom").
- Fixed the corner: the panel's tab strip is one row shorter than the editor's, so the outer border now turns at the strip's underline.
- While reproducing it I found the editor collapsing when docked right: the width clamp was reserving `MinTopH` (rows) as columns, so on narrower windows the editor got ~0 columns (it's also missing in your screenshot). It now keeps at least 20 columns.
- Rebased on main, added an e2e test for the layout, and documented `panel.position` in the settings reference.

eugenioenko (approved):

eugenioenko:
@SimonOcampo1 Thank you for adding this feature, with widescreens this make so much more sense. It's awesome! 

:tada:

PR #619, opened 2026-09-17
feat(workspace): browse for a folder instead of typing the whole path

SimonOcampo1:
Open Folder and Add Folder are a bare text input: you type the absolute path, or you do not open the folder. That is fine when you know where you are going and useless when you do not — there is no way to look.

This keeps the input and puts a browser under it.

```
┌─ Open Folder ──────────────────────────────┐
│ ❯ /home/simon/repos                        │
│ Home         ..                            │
│ Documents    braingent-sto                 │
│ Downloads    cv-creator                    │
│ Desktop      encyclopaedia-theologica      │
│ repos        …                             │
│                            Cancel    Open  │
└────────────────────────────────────────────┘
```

Places on the left, the current directory's subdirectories on the right, the way a file manager arranges them — in one list the shortcuts push the listing you are reading out of view.

## Where the places come from

The files GTK file managers already write:

- `~/.config/user-dirs.dirs` — the XDG directories
- `~/.config/gtk-3.0/bookmarks` — the user's own, `file:///path Name`, percent-escaped

Both are plain text, so this adds no dependency, and it reuses shortcuts the user has already curated somewhere else instead of asking them to curate a second list inside the editor. Bookmarks whose directory no longer exists are dropped — the file happily outlives the directory, and a dead shortcut is worse than none — and duplicates collapse.

The XDG names are localized (`Descargas`, `Escritorio`), which is exactly why the paths cannot be guessed from English defaults. Only the path comes from the file; the label shown is ours, so the list reads the same in any locale.

Nothing is read on platforms that do not have these files: `Places()` then returns just Home, and the dialog is still a browser.

## Interaction

The input stays the source of truth — typing a path is still the fastest route for anyone who knows it. The two sides stay in sync: selecting a row writes its path into the input, and typing re-lists as soon as the text names a real directory. A half-typed path deliberately leaves the listing where it was rather than blanking on every keystroke.

Tab moves between the columns, Enter descends, `Alt+O` or Enter in the input confirms. Subdirectories sort case-insensitively with dotfiles last.

## Tests

`internal/workspace`: both sources, missing directories, user dirs left unset (GTK marks those by pointing the entry at `$HOME`, and taking it literally lists Home several times), duplicates, percent-escapes, and stable ordering — the XDG entries were collected in a map at first, so the picker listed them differently on every run.

`internal/app`: listing order, the root case where there is no parent, and an unreadable directory still offering the way back out.

---

Happy to adjust the shape of this — the two-column split and putting it behind the existing commands rather than a new one are both judgement calls, and I would rather match what you had in mind than land it as-is.

## Summary by CodeRabbit

* **New Features**
  * Added a folder picker for selecting workspace directories without manually entering paths.
  * Added directory navigation, filtering, parent-folder access, and confirmation controls.
  * Added shortcuts for home, common user directories, and saved local bookmarks.
  * Workspace folder selection now uses the folder picker and validates directories before adding them.


eugenioenko:
@coderabbitai review

eugenioenko (approved):
Hi @SimonOcampo1, thank you for working on this feature and helping improving ttt for the community, really appreciate the effort, thank you :)

PR #620, opened 2026-09-18
fix(lsp): build file URIs that servers can actually parse

SimonOcampo1:
marksman crashes and takes the language server down for the rest of the session:

```
System.UriFormatException: Invalid URI: The hostname could not be parsed
    at Marksman.Paths.uriToSystemPath(String uri)
    at Marksman.Server.TextDocumentHover@826.Invoke(State state)
[lsp:markdown] server exited unexpectedly: EOF
```

## Cause

The URI is built by concatenation — `"file://" + path` — which is wrong in two ways, both of which come up in ordinary use:

- **A relative path turns its first segment into the authority.** `docs/a.md` becomes `file://docs/a.md`, where `docs` is the hostname. That is exactly what the message above is complaining about.
- **A path containing a space is not a valid URI.** `Ingenieria en Sistemas` is a real directory on the machine this was found on, and every document under it was announced with raw spaces in its URI.

Servers that parse URIs loosely had been hiding this. marksman does not, and it does not fail the request — it exits.

## Fix

`lsp.FileURI` and `lsp.URIToPath` resolve to an absolute path and escape it, and the three places that hand-built a URI now go through them: the document URIs, the root URI at `initialize`, and the one handed to the desktop file manager over dbus, which broke on spaces the same way.

```go
FileURI("docs/a.md")                 // file:///cwd/docs/a.md
FileURI("/home/u/a b/note.md")       // file:///home/u/a%20b/note.md
```

## Tests

The authority case, escaping, a round trip through paths with spaces, accents and a percent sign, and that a relative path comes back absolute — one reaching the server would make every position it reports refer to a different file.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of file paths containing spaces, Unicode characters, and special symbols.
  - Fixed file manager reveal actions to correctly open files across supported Linux environments.
  - Improved language server workspace path handling for more reliable project initialization.
  - Added more consistent conversion between local file paths and file URLs, including relative paths.


eugenioenko (approved):
@SimonOcampo1 thank you for the fix. I was seeing marksman errors often enough but didn't have time to look yet. Great catch! 

Thank you for making ttt better for the community, really appreciate it :)

PR #621, opened 2026-09-18
feat(markdown): give headings a visible hierarchy in the preview

SimonOcampo1:
Rendered markdown keeps the leading hashes and gives every heading the same bold style, so a level-1 and a level-4 look identical and the document reads as one flat wall of text. The preview shows the markup instead of the structure it describes.

Before / after on the same file:

```
# braingent-sto                     braingent-sto
                                    ══════════════════════════════
An orchestration layer over…        An orchestration layer over…

## Language                         Language
                                    ──────────────────────────────
Code, comments and docs are…        Code, comments and docs are…
```

## Change

The hashes are markup, not content, so they go. Levels 1 and 2 get a rule underneath instead — the usual way a terminal shows weight without relying on colour — drawn with a heavier character for level 1 so the two are told apart. Deeper levels stay bold, which is enough once the top of the tree is anchored.

The divider line now carries its rule character in its span rather than having the widget hardcode one. That also means a thematic break emits the rule character itself: it used to emit the literal `"---"`, which under the new rendering would paint repeated dashes instead of a line.

This affects everything the markdown widget renders, the plugin READMEs in the plugin detail pane included — those gain the same hierarchy.

## Tests

That no heading keeps a hash at any level, that levels 1 and 2 use different rules and level 3 gets none, and the existing divider test updated to assert the rule character rather than the markup that produced it.


eugenioenko (commented):
Hi @SimonOcampo1 ttt internal markdown is used for rendering lsp text which doesn't use headers. 

If you want a markdown preview,  you can use this plugin or create one based on it:
https://github.com/eugenioenko/ttt-plugins/tree/main/markdown-preview

SimonOcampo1:
Makes sense, thanks for the context. I didn't realize the internal renderer is only for LSP text. Closing this; the preview side lives in the plugin (eugenioenko/ttt-plugins#9).

PR #632, opened 2026-09-20
feat(terminal): highlight selected text with a themeable selection color

### LINKED ISSUE #631: Terminal selection highlight is washed out with transparent background
## Problem
With `editor.transparentBackground` on, selecting text in the integrated terminal produces a near-white highlight where the text is hard or impossible to read. With it off, the selection inverts correctly.

## Likely cause
Not reproduced under a debugger; this is from reading the code.

- The terminal selection is drawn by swapping each cell's foreground and background (`internal/ui/terminal_widget.go`, around lines 352 and 396). Cells with an unset color fall back to `Palette.Bg` for the new foreground and `Palette.Fg` for the new background.
- `BuildTerminalPalette` (`internal/app/theme.go:231`) leaves `Palette.Bg` unset when the background is transparent and the theme has no `terminal.background`.
- So for a cell with the default background, the swapped foreground stays unset and falls through to the terminal's own default foreground. The new background becomes the theme foreground. The result is light text on a light background.

The editor selection is not affected, since it uses `StyleSelection` with the reverse attribute and lets the host terminal invert.

## Expected
A readable selection with transparent background on, in both dark and light themes.

## Ideas
- Use the reverse attribute for the selected cells, as the editor does, so the host terminal does the inversion.
- Or, when `Palette.Bg` is unset, pick a contrasting foreground from the theme instead of the terminal default.

eugenioenko:
## Summary
- Adds `terminal.selection` to themes. It inherits `editor.selection.bg` when unset, the same way `fileIcons` inherit from the terminal colors.
- Selected terminal cells now take that background and keep their text color, matching how the editor renders selection.
- Fixes the washed-out selection with `transparentBackground` on: the old fg/bg swap had no palette background to swap in, so it produced terminal-default text on a light theme-foreground background. It also works over a wallpaper.
- Themes with no selection color at all still fall back to the previous fg/bg swap.
- Documented in the themes guide.

Fixes #631

## Test plan
- [x] `go test ./internal/config ./internal/ui ./internal/app`
- [x] Unit tests for inheritance, explicit override, and cell rendering with and without transparency
- [ ] Select text in the terminal with transparent background on, with and without a wallpaper
- [ ] Check contrast in the default theme, where `editor.selection.bg` is a subtle `#282828`

🤖 Generated with [Claude Code](https://claude.ai/code/session_019niStUGkMsHyhRyjDAiXeA)

https://claude.ai/code/session_019niStUGkMsHyhRyjDAiXeA

## Summary by CodeRabbit

* **New Features**
  * Added configurable terminal selection highlighting.
  * Terminal selections can use a dedicated theme color or inherit the editor selection background.
  * Selection highlighting now preserves terminal text colors when a selection background is configured.

* **Documentation**
  * Updated theme documentation and examples to describe the terminal selection color.


eugenioenko:
@coderabbitai review

